PR #13, opened 2026-08-10
feat(sequence): combined fragments (alt/opt/loop) + OAuth example

ActArtech:
## Summary

Sequence diagrams in this skill only documented linear happy paths. Real auth and API flows need combined fragments (`alt` / `opt` / `loop`) and a clear async open-arrowhead convention.

This PR expands the sequence type reference and adds a special **bearer + refresh** example trio (light / dark / full), without changing the existing cold-cache examples.

## Changes

- Expand `skills/diagram-design/references/type-sequence.md`
  - Combined fragment frames (`ALT` / `OPT` / `LOOP`)
  - Guard + divider primitives
  - Async open arrowhead (`arrow-open`)
  - Complexity budget and anti-patterns
- Add special examples:
  - `example-sequence-oauth.html`
  - `example-sequence-oauth-dark.html`
  - `example-sequence-oauth-full.html`
- Discoverability: README tree + quickstart note; `SKILL.md` §7 budget lines for fragments
- `scripts/verify-sequence-oauth.py` - structural checks + drives real `lint-skin.py`

## Non-goals

- `par` / `critical` / `break` / `ref` operators
- Participant create/destroy, found/lost messages
- Replacing or redesigning the cold-cache sequence trio

## Test plan

- [x] `python scripts/lint-skin.py skills/diagram-design/assets/example-sequence-oauth.html skills/diagram-design/assets/example-sequence-oauth-dark.html skills/diagram-design/assets/example-sequence-oauth-full.html` (0 findings)
- [x] `python scripts/lint-skin.py --all --baseline` (0 findings)
- [x] `python scripts/verify-sequence-oauth.py` (ALL GATES PASSED)
- [ ] Open `example-sequence-oauth.html` in a browser and sanity-check ALT frame + legend

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,8 @@ open ~/.claude/skills/diagram-design/assets/index.html
 # "Make me an architecture diagram of my app: frontend, backend, database, Redis cache."
 # "I need a quadrant showing Q2 projects by impact vs effort."
 # "Give me a sequence diagram of the OAuth handshake."
+# (branching token refresh uses the ALT combined-fragment grammar in type-sequence.md;
+#  see assets/example-sequence-oauth.html)
 ```
 
 Claude will pick the right type, build the HTML, and save it. You can also start from a template directly:
@@ -247,7 +249,8 @@ diagram-design/
 │   ├── template*.html               — scaffolds for new diagrams
 │   ├── example-<type>.html          — 3 variants × 27 types
 │   ├── example-loop-terminal.html   — terminal-variant flagship
-│   └── example-quadrant-consultant.html  — consultant-special 2×2 scenario matrix
+│   ├── example-quadrant-consultant.html  — consultant-special 2×2 scenario matrix
+│   └── example-sequence-oauth*.html — sequence special: bearer + ALT refresh
 └── docs/screenshots/                — the images in this README
 ```
 
--- a/scripts/verify-sequence-oauth.py
+++ b/scripts/verify-sequence-oauth.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python3
+"""Structural + skin-lint verification for sequence combined-fragment work.
+
+Drives the real shipped artifacts (type-sequence.md, oauth examples, lint-skin.py).
+Exit 0 only when all gates pass. Intended for local/PR checks alongside:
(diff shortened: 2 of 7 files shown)


SajidSabreen (changes requested):
Not a maintainer, just read through this. Nice work on the fragment grammar, and good call keeping par/critical/break/ref out of scope.

One thing that might be a problem, the returns aren't drawn consistently. M5 (200 · NEW ACCESS) is dashed like your Message kinds table says, but M3 (401) and M7 (the retry 200) are solid, which makes them look like calls. Neither matches anything in the diagram's own legend. Your inline comment on M5 says "return dashed, filled marker," so I'm guessing M3 and M7 just got missed. Same in all three variants.

the older Layout conventions bullet says returns match the color of the originating call, but the new table makes them muted. M5 follows the new rule and breaks the old one. Might be worth deleting the old bullet here so the file doesn't contradict itself.

Also, verify-sequence-oauth.py passed with all of the above in place, which suggests it's checking for strings rather than the layout rules the docs just introduced. It also duplicates the lint-skin.py run that's already in the Contributing section, and nothing runs it automatically. Might be simpler to drop it or move it to its own PR.

Small stuff
- README quickstart says "OAuth handshake" but this is a refresh flow. Legitimately OAuth, but the note will send people looking for the redirect and code exchange to the wrong diagram.
- First API activation bar ends at 268, but M2 fires at 272, so the return leaves a bar that just closed. h="104" would match how the other bars bracket their messages.
- The example has both an accent-stroked API box and a coral return, which hits SKILL.md's cap of 2 while the new doc says "prefer 1 for fragment diagrams." Do actor boxes count toward that?
- Frame inset is 72 left, 40 right. Looks intentional to keep [token valid] off the Client lifeline. If so, the ≥12px rule probably wants rephrasing as "clear the widest guard label."

cathrynlavery:
Thanks so much for this, @ActArtech — really solid addition. The combined fragment grammar (alt/opt/loop) was a genuine gap and the OAuth trio gives a great real-world reference for it. All three lint and verify gates passed clean. Appreciate the care you put into keeping the existing examples untouched and scoping the non-goals clearly. Welcome to contribute again anytime! 🙌

PR #18, opened 2026-08-11
feat(scripts): lint diagrams as rendered, not just as source

Yuvraj3905:
## What

  Adds `scripts/lint-render.py` — a companion to `lint-skin.py` that renders each example in headless Chromium and measures the laid-out result.

  `lint-skin.py` reads the HTML source. That catches off-skin colors and fonts, but it can't see anything that only exists after layout: a node nudged past the SVG viewport, an SVG that collapses to zero height, a page that scrolls sideways at export width, a runtime error. Those land silently in the gallery and in exported PNGs.

  ## Checks

  | Category | Fails when |
  |---|---|
  | `clipped` | an element's rect escapes its `<svg>` viewport (SVG clips there, so it's cut off in-browser and in PNG export) |
  | `svg-collapsed` | an `<svg>` renders at zero width or height |
  | `page-overflow` | the document scrolls horizontally at 1600px |
  | `page-error` / `console-error` | JS throws, or the console logs an error |

  Content inside `defs`, `clipPath`, `mask`, `marker`, `pattern`, `symbol`, gradients and filters is skipped (it's a definition, not painted where it sits), as are elements hidden via `display`, `visibility` or `opacity: 0`.

  ## Why invariants instead of golden images

  No PNGs enter the repo, there is nothing to re-record when a diagram legitimately changes, and there's no baseline hash to drift between machines or Chromium versions. A failure names the element and the overflow in pixels, so it's actionable without opening a diff viewer.

  ## Usage

  Mirrors `lint-skin.py`: same `--all` / `--quiet` flags, same `path: category: message` output, exit 1 on findings.

  ```bash
  pip install playwright && playwright install chromium   # same dep the PNG export already documents
  python3 scripts/lint-render.py --all
  python3 scripts/lint-render.py skills/diagram-design/assets/example-venn.html
  python3 scripts/lint-render.py --self-test
  ```

  Playwright missing → prints the install line from `references/export.md` and exits 2, matching how the export command behaves. Bundled Chromium, falling back to a system Chrome install.
                                           
  ## Verification
  - `--all` → **92 files rendered, 0 findings** on `main`. Clean baseline, so no skip-list file is needed.
  - `--self-test` → renders a deliberately broken diagram and asserts `clipped`, `svg-collapsed` and `page-overflow` all still fire. Guards against the lint silently becoming a no-op.
  - Real regression check: moved a painted circle in `example-venn.html` from `cx="500"` to `cx="1400"` → `clipped: circle escapes the svg viewport: right by 648.0px`, exit 1.
  - Missing-Playwright path → exit 2 with the install instruction.

  Remote webfont request failures are deliberately not reported — offline and proxied machines just get fallback metrics, which isn't the diagram's fault.

  ## Not verified

  Only tested on one Linux box with bundled Chromium; the `channel="chrome"` fallback never had to trigger. No CI workflow added — say the word and I'll wire it into a GitHub Action alongside `lint-skin.py`.

  ## Scope

  One new script, plus a README paragraph under *Contributing / skin lint*. No changes to any diagram, template, reference file or existing script.


rajpratham1 (approved):
From the PR you shared, the changes mainly add compatibility metadata to the MiniMax skills and update the localized READMEs. If those changes look correct and you don't see a bug or unwanted behavior, Approve is appropriate.

cathrynlavery:
Thanks for the thoughtful proposal. Render-level validation would be valuable here, but we cannot merge the current implementation because the core measurement can produce both false confidence and noisy failures.

The main changes needed are:

1. Replace or narrow the `getBoundingClientRect()` painted-bounds model. Adversarial Chromium checks showed false negatives for clipped thick strokes, markers, and filter effects, plus false positives for safe `clipPath` geometry, content hidden by an ancestor, and SVGs using `overflow: visible`.
2. Fix font readiness. `Page.evaluate(..., timeout=5000)` is not a valid Python Playwright call, so it raises `TypeError` and the broad exception handler silently skips the wait.
3. Do not suppress every `Failed to load resource` console error; that also hides missing local assets. Filter only the intended remote-font failures.
4. Define the trust boundary for loading contributor HTML with JavaScript and network access, or disable/isolate those capabilities.
5. Add negative/adversarial browser tests and CI coverage before describing this as a mandatory repository linter.

The current self-test passes and all 92 examples report clean on macOS Chromium, but those happy-path results do not exercise the failure modes above. We would be happy to re-review a revision that addresses the painted-bounds model and font wait first; otherwise we will not take this version.

Yuvraj3905:
Hey @cathrynlavery , I did the changes u recommended can you please verify.

philyuchkoff (commented):
This is a genuinely thoughtful design, @Yuvraj3905 — the paint-as-oracle approach (screenshot as authored vs. overflow released, diff in-browser) is the right call for the clipping problem, and the self-test suite with must-NOT-flag cases is what separates a reliable linter from a cry-wolf one. The network block and the  category are both good trust-boundary and honesty decisions. I verified the concept locally against the repo's 94 examples: clean run, no false positives — exactly what a gate should do.

A few things to resolve before merge:

## 1. Path conflict with PR #27 — must coordinate

PR #27 ("add robust Playwright rendered diagram layout validator scripts/lint-render.py") adds the *same* file path `scripts/lint-render.py`; the two can't both merge. The #27 author credits this PR as the origin and positions his as a fix of the maintainer's review findings — but the implementations are incompatible (geometry-based vs. paint-based), so one has to win. Given the maintainer review on #18 seems to point toward this direction, I'd suggest: close #27 and fold its genuinely useful bits (fixture deduplication, CI wiring ideas) into this PR, or have the authors agree on a single implementation. Whichever way, decide before merge.

## 2. The new workflow duplicates an existing gate

`.github/workflows/lint.yml` re-runs `lint-skin.py --all --baseline` in its `skin` job — that gate already runs in `.github/workflows/ci.yml`. The repo validates everything through one workflow; better to add the render steps to `ci.yml` instead of a parallel workflow, so the gate stays in one place and CI minutes stay low.

## 3. Silent Chrome fallback masks environment issues

`launch()` falls back to `channel="chrome"` on any exception. In CI there is no system Chrome, so a real Playwright problem (bad install, version mismatch) would surface as a confusing fallback failure rather than a clear "Playwright Chromium missing" message. At minimum, log a warning when falling back; consider making the fallback opt-in (`--system-chrome`).

## 4. Font metrics in CI differ from local runs — document it

Network is blocked in CI, so Geist/Instrument come from fallback stacks, and `--fonts` isn't exercised there. That's the right tradeoff for reproducibility, but text metrics (and thus borderline overflow findings) can differ between CI and a dev machine with `--fonts`. A sentence in README or the docstring clarifying "CI measures with fallback fonts by default" would prevent confusion when a contributor sees a finding locally that CI doesn't (or vice versa).

## Minor

- `page.screenshot(animations="disabled")` — fine; maybe assert `--all` never silently skips an `unmeasurable` svg (notes shouldn't become an accidental escape hatch over time).
- Consider emitting findings sorted by file/category for stable output, matching `lint-skin.py` conventions.

The core is strong — once the coordination question and the workflow dedup are sorted, this is a merge-ready gate. Happy to help verify once it's rebased.

philyuchkoff (commented):
This is a genuinely thoughtful design — the paint-as-oracle approach (screenshot as authored vs. overflow released, diff in-browser) is the right call for the clipping problem, and the self-test suite with must-NOT-flag cases is what separates a reliable linter from a cry-wolf one. The network block and the unmeasurable category are both good trust-boundary and honesty decisions. I verified the concept locally against the repo's 94 examples: clean run, no false positives — exactly what a gate should do.

A few things to resolve before merge:

## 1. Path conflict with PR #27 — must coordinate

PR #27 ("add robust Playwright rendered diagram layout validator scripts/lint-render.py") adds the *same* file path `scripts/lint-render.py`; the two can't both merge. The #27 author credits this PR as the origin and positions his as a fix of the maintainer's review findings — but the implementations are incompatible (geometry-based vs. paint-based), so one has to win. Given the maintainer review on #18 seems to point toward this direction, I'd suggest: close #27 and fold its genuinely useful bits (fixture deduplication, CI wiring ideas) into this PR, or have the authors agree on a single implementation. Whichever way, decide before merge.

## 2. The new workflow duplicates an existing gate

`.github/workflows/lint.yml` re-runs `lint-skin.py --all --baseline` in its skin job — that gate already runs in `.github/workflows/ci.yml`. The repo validates everything through one workflow; better to add the render steps to `ci.yml` instead of a parallel workflow, so the gate stays in one place and CI minutes stay low.

## 3. Silent Chrome fallback masks environment issues

`launch()` falls back to `channel="chrome"` on any exception. In CI there is no system Chrome, so a real Playwright problem (bad install, version mismatch) would surface as a confusing fallback failure rather than a clear "Playwright Chromium missing" message. At minimum, log a warning when falling back; consider making the fallback opt-in (`--system-chrome`).

## 4. Font metrics in CI differ from local runs — document it

Network is blocked in CI, so Geist/Instrument come from fallback stacks, and `--fonts` isn't exercised there. That's the right tradeoff for reproducibility, but text metrics (and thus borderline overflow findings) can differ between CI and a dev machine with `--fonts`. A sentence in README or the docstring clarifying "CI measures with fallback fonts by default" would prevent confusion when a contributor sees a finding locally that CI doesn't (or vice versa).

## Minor

- `page.screenshot(animations="disabled")` — fine; maybe assert `--all` never silently turns unmeasurable notes into an accidental escape hatch over time.
- Consider emitting findings sorted by file/category for stable output, matching `lint-skin.py` conventions.

The core is strong — once the coordination question and the workflow dedup are sorted, this is a merge-ready gate.

philyuchkoff:
✖ Duplicate note: the first review from philyuchkoff was posted by mistake (a broken terminal command). The actual full review is in the next review comment below.

Yuvraj3905:
> This is a genuinely thoughtful design — the paint-as-oracle approach (screenshot as authored vs. overflow released, diff in-browser) is the right call for the clipping problem, and the self-test suite with must-NOT-flag cases is what separates a reliable linter from a cry-wolf one. The network block and the unmeasurable category are both good trust-boundary and honesty decisions. I verified the concept locally against the repo's 94 examples: clean run, no false positives — exactly what a gate should do.
> 
> A few things to resolve before merge:
> 
> ## 1. Path conflict with PR #27 — must coordinate
> PR #27 ("add robust Playwright rendered diagram layout validator scripts/lint-render.py") adds the _same_ file path `scripts/lint-render.py`; the two can't both merge. The #27 author credits this PR as the origin and positions his as a fix of the maintainer's review findings — but the implementations are incompatible (geometry-based vs. paint-based), so one has to win. Given the maintainer review on #18 seems to point toward this direction, I'd suggest: close #27 and fold its genuinely useful bits (fixture deduplication, CI wiring ideas) into this PR, or have the authors agree on a single implementation. Whichever way, decide before merge.
> 
> ## 2. The new workflow duplicates an existing gate
> `.github/workflows/lint.yml` re-runs `lint-skin.py --all --baseline` in its skin job — that gate already runs in `.github/workflows/ci.yml`. The repo validates everything through one workflow; better to add the render steps to `ci.yml` instead of a parallel workflow, so the gate stays in one place and CI minutes stay low.
> 
> ## 3. Silent Chrome fallback masks environment issues
> `launch()` falls back to `channel="chrome"` on any exception. In CI there is no system Chrome, so a real Playwright problem (bad install, version mismatch) would surface as a confusing fallback failure rather than a clear "Playwright Chromium missing" message. At minimum, log a warning when falling back; consider making the fallback opt-in (`--system-chrome`).
> 
> ## 4. Font metrics in CI differ from local runs — document it
> Network is blocked in CI, so Geist/Instrument come from fallback stacks, and `--fonts` isn't exercised there. That's the right tradeoff for reproducibility, but text metrics (and thus borderline overflow findings) can differ between CI and a dev machine with `--fonts`. A sentence in README or the docstring clarifying "CI measures with fallback fonts by default" would prevent confusion when a contributor sees a finding locally that CI doesn't (or vice versa).
> 
> ## Minor
> * `page.screenshot(animations="disabled")` — fine; maybe assert `--all` never silently turns unmeasurable notes into an accidental escape hatch over time.
> * Consider emitting findings sorted by file/category for stable output, matching `lint-skin.py` conventions.
> 
> The core is strong — once the coordination question and the workflow dedup are sorted, this is a merge-ready gate.

Rebased onto main (`4da4dfb`) and addressed both reviews.
 
  **Workflow duplication** — `lint.yml` is gone. The render steps now live in the existing `.github/workflows/ci.yml`, in the `validate` job, before the icon-assets check:

  ```yaml
        - name: Install Playwright Chromium
          run: |
            pip install playwright
            playwright install --with-deps chromium

        # The checks before the sweep: a broken linter must not be able to report
        # a clean repository.
        - name: Verify render linter checks
          run: python3 scripts/lint-render.py --self-test

        - name: Run render linter
          run: python3 scripts/lint-render.py --all
  ```

  **Silent Chrome fallback** — removed. A missing bundled Chromium now fails loudly with the install command; using a system browser is opt-in via `DIAGRAM_LINT_BROWSER_CHANNEL=chrome`, which prints the channel it used to stderr. In CI, a broken environment stays visible.

  **Stable output** — findings are sorted within each file; the file list was already sorted.

  **Font metrics in CI vs local** — documented in the README. Network is blocked by default, which is what CI runs, so text is laid out in the fallback faces: deterministic and machine-independent, but not what a reader sees. `--fonts` allows  `fonts.googleapis.com` and `fonts.gstatic.com` and nothing else, for when you want to check whether real text fits its box.

  **Verified on the rebased branch:** `--self-test` 15/15; `--all` 94 files, 0 findings, 0 notes; and the rest of the CI gates still pass locally (`test-lint-a11y.py`, `lint-skin.py --all --baseline`, `verify-sequence-oauth.py`, `verify-drawio-import.py`, `verify-mermaid-import.py`).

  ---

  **On the path conflict with #27** — happy to defer to whichever you prefer, but since both PRs claim `scripts/lint-render.py` and the difference is the *measurement model*, it's worth naming rather than settling by merge order.

  #27 computes `rect.right - svgRect.right` from `getBoundingClientRect()` with a 2px tolerance. I probed that model in Chromium while fixing the first review here, and it fails in both directions:

  - A `stroke-width: 40` rect in a 200px viewport reports `right=195` — inside — while ink paints to 215. Arrowhead markers and filter bleed are equally invisible, because `getBoundingClientRect()` on an SVG child returns geometry, not paint.
  `getBBox({stroke: true, markers: true})` isn't implemented in Chromium either, so there's no cheap fix.
  - Its hidden-element filter reads each element's *own* computed style, so children of an `opacity: 0` ancestor still flag (opacity doesn't inherit). `clip-path` that pulls geometry back inside also still flags.

  Those are exactly the false negatives and false positives that blocked my first version, so I replaced geometry with paint: screenshot as authored, screenshot with the SVG's `overflow` released, diff the two. Ink appearing outside was being cut off.
  `--self-test` asserts this on 15 fixtures, and roughly half are cases that must **not** be flagged — `clip-path`-safe, `opacity: 0` ancestor, `display: none` ancestor, rounded `overflow: hidden` ancestor, `overflow: visible` SVG, fully-inside content.
  Two of those guards exist because this repo's own files false-flagged during development: `example-loop-terminal.html` (its `border-radius` wrapper loses rounded corners when released) and the `high-level*` / `datalake-full` files (1px boundary re-antialiasing).

  Two things #27 does better that I'd like either way: writing a screenshot artifact on failure, and keeping fixtures in `scripts/fixtures/` rather than inline strings. Glad to port both into this PR, or to hand the paint-based measurement to @parveen0029 for #27 — whichever gets you one linter instead of two.

parveen0029:
Hi @Yuvraj3905 👋

Thank you so much for the thoughtful breakdown, detailed Chromium probing, and compliments on the screenshot artifact engine and fixture separation! I would be genuinely happy for you to lead and drive PR #18 to merge as the unified rendered linter gate.

Since #18 was the original PR that pioneered this effort, it makes total sense to fold the useful pieces from #27 into #18 so maintainers get one rock-solid linter instead of competing PRs.

### 💡 A Few Research Notes & Suggestions for the Paint Diffing Algorithm

While testing paint-based diffing across cross-platform environments, here are a few subtle edge cases you might find helpful to guard against:

1. **Animation / Transition Freezing**: Injecting `* { animation: none !important; transition: none !important; }` (or setting Playwright `animations: "disabled"`) before capturing Screenshot 1 vs Screenshot 2 prevents animated SVGs or CSS keyframes from triggering false-positive pixel diffs.
2. **Targeting `<svg>` vs Wrapper Containers**: Scoping `overflow: visible !important` strictly to `<svg>` elements rather than ancestor card containers ensures rounded corners (`border-radius: 8px`) on outer wrappers don't unclip and produce false corner diffs.
3. **Subpixel Antialiasing Tolerance**: Adding a small pixel tolerance threshold (1–2px) prevents cross-OS font rendering variations (Windows vs Linux font smoothing near boundaries) from flagging as overflow.
4. **Porting #27 Screenshot Artifacts**: You can port the `--screenshot-dir` CLI option and `scripts/fixtures/sample-broken-layout.html` from #27 so that failure PNG screenshots get automatically zipped and uploaded to GitHub Actions artifacts (`upload-artifact@v4`).

Feel free to cherry-pick or port any code from #27 into #18! I will go ahead and close #27 to clear the path for #18 to merge cleanly.

Thanks again for the great collaboration and leadership on this!

cathrynlavery:
Thank you, @Yuvraj3905 — I really appreciate the depth of thought behind the paint-diff approach, the self-test suite, and your collaboration with @parveen0029 to converge on one rendered-layout linter. Detecting what the browser actually paints is the right problem to solve.

A deeper adversarial run found several blockers before this can safely become a pull-request gate:

- Contributor HTML can bypass the current request routing with a WebSocket, so the documented network isolation is not yet true. Please explicitly block WebSockets and self-test against a local listener; font exceptions should require HTTPS plus an exact hostname.
- The measurement pass clears authored inline transforms and restores overflow to an empty value rather than its exact original value. Snapshot and restore every mutated property in `finally`, and assert the DOM/style state is unchanged afterward.
- SVGs inside common clipping wrappers can pass clean because `overflow:visible` returns early and the released-ancestor rectangle masks changed paint. Please stage root and ancestor releases separately and add class-based/inline wrapper fixtures.
- A scrolled-into-view SVG mixes viewport and document coordinates, which can hide top clipping. Keep one viewport coordinate system and add paired above/below-fold tests.
- Pin Playwright/browser behavior for a pixel-based CI oracle rather than installing whatever is newest on every run.

The current self-test (15/15), all 94 examples, and the full repository suite pass locally after integrating current main. These are adversarial coverage gaps, not a dismissal of the strong foundation. Thank you again — this is valuable work, and I want the security boundary and paint oracle to be as trustworthy as the idea deserves.

Yuvraj3905:
@cathrynlavery 
All five reproduced in Chromium before changing anything — every one was real, and two were worse than reported.

  **1. WebSocket bypass.** Confirmed: with request routing alone, a page pointed at a local listener opened a real TCP connection to it. Network is now cut at the browser's resolver (`--host-resolver-rules=MAP * ~NOTFOUND`), which covers WebSockets, EventSource and anything else that opens its own socket, with request routing kept as a second layer. `--fonts` appends `EXCLUDE fonts.googleapis.com,EXCLUDE fonts.gstatic.com` and the route handler now requires `scheme == "https"` and an exact `hostname` match — the old check was a substring test, so a lookalike host or a path containing the string would have passed.

  **2. Style restoration.** Correct — `transform` was reset to `''`, destroying an authored inline transform rather than restoring it. Every mutation now snapshots the element's exact `style` attribute; restore puts it back verbatim (or removes the attribute if there wasn't one), in reverse order, from a `finally`. A fixture carries `transform: rotate(2deg)`, and the self-test asserts `documentElement.outerHTML` is byte-identical before and after measuring.

  **3. Wrapper masking.** Both halves confirmed. Releases are now staged: stage 0 releases the SVG alone and looks for ink outside the SVG's own box; stage *k* also releases the *k*  nearest clipping ancestors and looks for ink outside the outermost released box. `overflow: visible` on the SVG now skips only stage 0 instead of returning early, so a wrapper that clips an SVG which doesn't clip itself is still checked. New fixtures: inline wrapper, class-based wrapper, and `overflow: visible` SVG inside a clipping wrapper.

  **4. Coordinate mixing.** Worse than described. For an SVG below the fold the old code computed `interior.top = bounding_box.y - scrollY = -800` where the correct value was `500`  so the interior box landed off-screen entirely. Everything is now one coordinate system: viewport pixels straight from `getBoundingClientRect()`, matching `page.screenshot()`. `bounding_box()`, `scrollX` and `scrollY` are gone. Paired above/below-fold spill fixtures plus a below-fold clean case. This never showed up in my sweeps because every example in the repo sits above the fold — the check was passing for the wrong reason.

  **5. Pinned oracle.** `PLAYWRIGHT_VERSION: "1.62.0"` in `ci.yml`, with the Chromium build chosen by that pin and the resolved version echoed into the build log. Each run also prints the browser version and the active resolver rule to stderr, so a pixel result is always traceable to the browser that produced it.

  **The guards aren't vacuous** — I mutation-tested the two new ones:

  - resolver block disabled → `network-isolation: 1 connection(s) reached a local listener`
  - restore disabled → `dom-restore: markup changed after measuring`, plus both wrapper cases collapse

  **Current state:** self-test 23/23 (over half are cases that must *not* be flagged); `--all` 94 files, 0 findings, 0 notes; injected regression still caught (`right by 652px`, exit
  1); `--fonts` clean; the other five CI gates pass locally.

  Remaining ceilings, unchanged and documented in the docstring: spill of ~2px or less (the antialiasing guard band), spill beyond 0.25× framing, and SVGs inside a *scrolling* ancestor those report `unmeasurable`, which prints but does not fail. Still Linux Chromium only on my side; the thresholds are tuned against this repo's examples on one machine, so a macOS run may want them revisited.

cathrynlavery (changes requested):
Thank you for the thorough revision. I validated exact head `6cd276446919139def140652513a6598fe067ecc`: the render-linter self-test passes all 23 cases and all 94 examples render with 0 findings. The branch is now behind current `main`, though, and cannot satisfy the repository's current merge gates: several required verifier scripts are absent, strict plugin validation fails on the old package metadata, there is no synchronized plugin-version bump, and there are no current-head CI results. Please rebase onto current `main`, bump both plugin manifests to a version greater than the new base, run the complete CONTRIBUTING.md suite, and obtain current-head CI with all threads resolved. Once that is updated, we can re-review the paint-based oracle on the current package.

Vitanakpapa:
Yuvraj3905:feat/render-lint

Yuvraj3905:
@cathrynlavery , Hi can you please recheck.

cathrynlavery:
The latest head appears to address the prior structural request: it merged current main, synchronized 2.4.1, and added a pinned Linux/Chromium CI leg. I’m sequencing #75 first because both PRs claim 2.4.1. After #75 lands, please rebase/bump again and obtain current-head CI; I’ll then re-review the paint oracle and can resolve the earlier change request if it remains clean.

cathrynlavery:

Yuvraj3905:
Hi @cathrynlavery , can you please approve the workflow so that I can make a final commit and after that this PR can be merged.

cathrynlavery:
Approved. if you can resolve conflicts then we can merge

Yuvraj3905:
Hi @cathrynlavery , resolved the commits. Ready to go!

cathrynlavery:

cathrynlavery:

cathrynlavery:
@Yuvraj3905, thank you for building the rendered-layout linter. Rendering the actual diagrams in Chromium closes an important gap that source checks cannot see, and the work is already paying off—the same browser-based approach helped catch the mobile gallery regression during this release. I really appreciate the care you put into the isolation, clipping checks, and test coverage.

PR #24, opened 2026-08-12
feat(import): redraw Mermaid diagrams at a chosen format, size, and detail level

mvanhorn:
## Summary

- add `skills/diagram-design/scripts/mermaid_extract.py`, a stdlib-only extractor that emits the same digest shape `drawio_extract.py` produces
- add the `/diagram-design:import-mermaid` command, `references/import-mermaid.md`, three fixtures, and a worked example
- accept `.mmd`, `.mermaid`, and Markdown containing one or more fenced `mermaid` blocks, with the same `--diagram` selector semantics as a multi-page `.drawio`
- add `scripts/verify-mermaid-import.py`, mirroring `verify-drawio-import.py`

## Why

#22 built the whole redraw pipeline one day ago: the four output dials in `references/output-spec.md`, the redraw-not-convert procedure, the fidelity ledger, and the taste gate. None of it is draw.io specific except the extractor. Every step after Step 1 consumes only the digest, so a Mermaid extractor that emits the same digest reuses the pipeline unchanged, and `import-mermaid.md` Steps 2-6 stay near-identical prose to the draw.io reference.

Two digest fields differ, and the reference documents why. There is no `source canvas` line, because Mermaid has no coordinates. That is convenient here: the "discard source coordinates" step is vacuous, so the digest prints `- source layout: none (Mermaid is layout-free); direction: TD` and carries the declared direction as a layout hint instead. Node shapes are normalized to the existing shape-family vocabulary `classify_shape` already uses, so the Step 3 type-selection table stays valid.

Four grammars parse: `flowchart`/`graph`, `sequenceDiagram`, `stateDiagram-v2`, and `erDiagram`. The parser is line-oriented recursive descent, not a regex soup and not a vendored copy of Mermaid's grammar. The cases a reviewer will probe are handled explicitly: `%%` comments and `%%{init}%%` directives, labels containing `-->` or brackets, `<br/>` normalization, markdown-string labels, HTML entities, `subgraph` nesting, inline `:::class` attachments, YAML frontmatter, semicolon-separated statements, and the labeled dotted and thick link forms (`A -. retry .-> B`, `A == critical ==> B`).

Source styling is discarded on purpose, matching the draw.io path: init themes, `classDef`, `class`, `:::class`, `linkStyle`, and frontmatter config are all counted and dropped. The extractor parses bounded text and never renders, evaluates, fetches, or follows a click target. Labels reach the digest escaped, so a label carrying encoded HTML stays inert when the digest is written with `--out` and rendered somewhere else.

## Evidence

Mermaid import and redraw

The fixture README, the digest, the redraw, and the gate. Simulated demo, rendered from the real command output above.

## Verification

- `python3 scripts/verify-mermaid-import.py`
- `python3 scripts/verify-drawio-import.py`
- `python3 scripts/verify-sequence-oauth.py`
- `python3 scripts/lint-skin.py --all --baseline`
- browser-rendered `example-import-mermaid.html`: one SVG, no overflow, no page errors


cathrynlavery:
Thank you, Matt, for building out Mermaid import and for the unusually thorough fixtures, trust-boundary work, and verification suite. The feature direction is valuable; the remaining work is compatibility with current Mermaid syntax before it can merge.

PR #23, opened 2026-08-12
ci: add GitHub Actions workflow for automated linting and verification gates

parveen0029:
Hi and Namaste @cathrynlavery 👋

Adds a native GitHub Actions CI workflow (`.github/workflows/ci.yml`) to automatically validate incoming pull requests and commits against all repository verification gates.

### Summary & Motivation
As noted in recent PR reviews, having an automated CI check on GitHub ensures that every new pull request is automatically validated against the skin linter and verification scripts without requiring manual local runs.

### What's Included
- `.github/workflows/ci.yml`: Native GitHub Actions workflow running on `ubuntu-latest` with Python 3.11, triggered on `push` and `pull_request` to `main`.
- Automated validation steps:
  1. `python scripts/lint-skin.py --all --baseline` (Skin & density linter)
  2. `python scripts/verify-sequence-oauth.py` (Sequence diagram specs)
  3. `python scripts/verify-drawio-import.py` (Draw.io import extractor)
  4. `python scripts/build-icons.py` (Icon build script & SVG asset integrity)
- Updated `README.md` under `Contributing / skin lint` documenting the automated CI check.

### Verification
- Executed all 4 verification scripts locally — all linter gates pass green cleanly.

PR #25, opened 2026-08-12
feat(a11y): make the accessible-SVG pattern a contract

mvanhorn:
## Summary

- add `role="img"`, a resolving `aria-labelledby`, and first-child `<title>` / `<desc>` slots to `template.html`, `template-dark.html`, `template-full.html`, and `template-terminal.html`
- add an `a11y` category to `scripts/lint-skin.py` that rejects a diagram SVG without a resolvable accessible name
- add `scripts/test-lint-a11y.py`, covering nine rejection cases and seven acceptance cases
- emit `aria-hidden="true"` from `scripts/build-icons.py` so decorative specimen icons stay exempt through a regeneration
- document the contract in SKILL.md section 10, `references/output-spec.md`, `references/export.md`, and the README

## Why

The pattern was already written by hand in the most recent work:

```html
<svg viewBox="0 0 1040 620" role="img" aria-labelledby="title desc">
<title id="title">Agent team org chart</title>
<desc id="desc">Org chart showing a command center routing work to specialist agents and escalation owners.</desc>
```

It was not written down anywhere, so it is absent from most shipped output. Measured on `main` at `da45d4a`: 22 of 99 files in `assets/` carry `role="img"` plus `<title>` plus `<desc>`, and 0 of the 4 templates do. The template number is the one that matters, because SKILL.md section 10 tells the agent to build every new diagram from those templates. A pattern that lives only in one hand-written example is a pattern the next diagram will not have.

Nine checks make the contract enforceable rather than aspirational. `role="img"` is required. `aria-labelledby` must be present, must resolve, and must name the `<title>` and `<desc>` IDs. `<title>` must be the first child element. `<desc>` must be non-empty. The accessible-name IDs must not be bare `title` / `desc`, must not still hold the template placeholder, must not be duplicated within one document, and must match the diagram slug.

The slug rule is what makes multiple exported SVGs safe to inline in one page: two files both using `title` collide, and the second figure resolves to the first figure's accessible name. Rejecting bare IDs alone does not close that, since two files could both pick `fixture-title`. The linter derives the expected slug from the filename, so `example-venn-dark.html` must use `venn-dark-title` and `venn-dark-desc`.

The templates ship a `[diagram-slug]` placeholder, and the linter rejects an unfilled placeholder, so the template itself cannot be published as a finished diagram.

`build-icons.py` is included because the generator, not the file, is the source of truth for `icons.html` and `references/primitive-icons.md`. Patching those two artifacts alone would have been undone by the next regeneration, and the documented icon-copying workflow would keep producing nested `<svg>` elements the new linter rejects. Regenerating after the change produced a byte-identical `icons.html`.

## Evidence

Accessible-SVG contract

The gap, the missing rule, the contract in the template, and the enforced suite. Simulated demo, rendered from the real command output above.

## Verification

- `python3 scripts/test-lint-a11y.py`
- `python3 scripts/lint-skin.py --all --baseline`
- `python3 scripts/verify-drawio-import.py`
- `python3 scripts/verify-sequence-oauth.py`
- regenerated `icons.html` and `references/primitive-icons.md` with `scripts/build-icons.py`


cathrynlavery:
@greptile-apps review

cathrynlavery (approved):
Maintainer verification passed locally: scripts/test-lint-a11y.py, lint-skin --all --baseline, verify-drawio-import.py, and verify-sequence-oauth.py are all green. Approved pending the repository Greptile 5/5 gate and CI integration after #23 lands.

cathrynlavery:
Thank you, Matt, for turning the accessible-SVG pattern into a documented, enforceable contract and carrying it through the templates, examples, generator, and adversarial tests. This is a substantial accessibility improvement.

PR #30, opened 2026-08-12
fix: repair mojibake arrow directions

Mr-Neutr0n:
## Summary

`scripts/fix-mojibake.py` had the same corrupted up-arrow key twice. The first entry rewrote up arrows to left arrows, and the actual left-arrow corruption sequence was absent, so both directions were repaired incorrectly.

This change uses the correct CP1252 undefined-byte sequence for left arrows, keeps the up-arrow mapping distinct, and adds a focused regression test to CI.

## Verification

- `python3 scripts/test-fix-mojibake.py`
- `python3 -m py_compile scripts/fix-mojibake.py scripts/test-fix-mojibake.py`
- `git diff --check`
- Repository preflight, secret, burst, and title gates pass.

The repository CI continues to run the full Python 3.11 validation matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Verify accessible SVG contract
         run: python3 scripts/test-lint-a11y.py
 
+      - name: Verify mojibake repair utility
+        run: python3 scripts/test-fix-mojibake.py
+
       - name: Run skin linter
         run: python3 scripts/lint-skin.py --all --baseline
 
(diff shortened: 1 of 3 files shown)


cathrynlavery (approved):
The arrow mappings are corrected, the focused regression test covers all four directions, and the complete local and repository checks pass. Thank you for catching this subtle data-repair bug.

cathrynlavery:
Thank you, @Mr-Neutr0n — I really appreciate this precise fix. The duplicated mojibake key was easy to miss, and adding a focused four-direction regression test makes the repair utility materially safer.

PR #31, opened 2026-08-12
fix(icons): preserve Devicon source viewBox

Mr-Neutr0n:
## What changed

`normalize_devicon()` now carries through the source SVG viewBox instead of assuming every Devicon uses 128×128. This keeps 32×32 assets inside their intended coordinate system.

## Verification

- Added `scripts/test-build-icons-devicon.py` with a 32×32 regression fixture
- Added the focused check to CI
- Ran all repository verification scripts and `lint-skin.py --all --baseline`

The commit is DCO signed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Verify accessible SVG contract
         run: python3 scripts/test-lint-a11y.py
 
+      - name: Verify Devicon normalization
+        run: python3 scripts/test-build-icons-devicon.py
+
       - name: Run skin linter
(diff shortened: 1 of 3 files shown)


cathrynlavery (approved):
The source viewBox is preserved correctly, its regression test passes, current main is integrated with all CI checks retained, and the repository checks are green.

cathrynlavery:
Thank you again, @Mr-Neutr0n — preserving each Devicon’s source viewBox fixes a subtle rendering assumption and makes the icon pipeline more faithful to upstream assets. I especially appreciate the focused 32×32 regression fixture.

PR #32, opened 2026-08-12
fix(icons): normalize quoted URL SVG attributes

Mr-Neutr0n:
## What changed

The generic URL SVG normalizer now accepts both single- and double-quoted attributes for `viewBox`, `fill`, and `stroke`. Previously, valid single-quoted SVGs could keep source colors and use the fallback canvas.

## Verification

- Added `scripts/test-build-icons-quoted-attributes.py` with a single-quoted SVG fixture
- Added the focused check to CI
- Ran all repository verification scripts and `lint-skin.py --all --baseline`

The commit is DCO signed.


rajpratham1 (approved):
The change has a clear purpose: it makes the SVG normalization logic handle both single-quoted and double-quoted attributes, and it adds a regression test specifically for that case.

The test also checks the important expected behavior:

viewBox is preserved correctly.
Original colors are normalized.
fill becomes currentColor.
stroke becomes currentColor.
The new test is added to CI, so the regression should be caught in future changes.

I don't see a clear blocking bug in the shown diff.

cathrynlavery (approved):
Single- and double-quoted URL SVG attributes are normalized correctly. The revision preserves unrelated inline styles, tests the path directly, refreshes generated assets, integrates current main, and passes all checks.

cathrynlavery:
Thank you, @Mr-Neutr0n — I really appreciate this third careful icon-pipeline contribution. Supporting valid single-quoted SVG attributes closes a real normalization gap. I added a separate follow-up commit so inline stroke, opacity, and other style declarations are preserved while colors are normalized, and expanded the regression test around your original fixture.

PR #52, opened 2026-08-13
fix(readme): serve right-sized previews so they stop failing to load

### LINKED ISSUE #77: [Bug]: README pictures don't load.
### Area

Documentation

### What happened?

Maybe a known bug but most of the diagram links on the repo README seem to be broken.

Image

### Steps to reproduce

Open the README File on Github.

### Environment

_No response_

### Validation gates

- [ ] `python3 scripts/test-lint-a11y.py`
- [ ] `python3 scripts/lint-skin.py --all --baseline`
- [ ] `python3 scripts/verify-drawio-import.py`
- [ ] `python3 scripts/verify-mermaid-import.py`
- [ ] I haven't run any gates

### Relevant output

```text

```

alexwickstrom:
## What does this PR do?

The README grid renders 41 screenshots at 252 CSS px, but every `` pointed at the full-resolution original (2400–2800px wide, **~5.0 MB** across the page). This adds `scripts/build-readme-thumbs.py`, commits right-sized WebP previews to `docs/screenshots/thumbs/`, and points README.md at those — **4998 KB → 336 KB**. The originals are untouched and every preview links to its own, so full detail stays one click away.

Fixes #77

### Why

Previews intermittently fail to load and paint blank — no broken-image icon, no alt text, a different subset each time. A reload fixes it, because the second load is a cache hit. Two mechanisms, both from the same cause:

1. **In-flight stalls.** 41 oversized requests go out at once over one HTTP/2 connection; bandwidth splits 41 ways and the slowest streams can be reset. A reset image that never *errored* renders as an empty box, which is why there's no broken-image icon.
2. **Decode memory.** 41 × ~5 MP × 4 bytes ≈ **820 MB of decoded bitmap** for one page of thumbnails, well past what a browser will hold; discarded decodes repaint blank.

Reported on Safari; two screenshots of the same section showed different cells blank, which is the signature of a load race rather than a bad file. All files are valid and return 200; nothing is broken server-side.

### What changed

| | before | after |
|---|---|---|
| README preview bytes | 4998 KB | **336 KB** |
| grid previews | 2400–2800px PNG | 600px WebP q82 |
| inline figures | 2400–2800px PNG | 1400px WebP q90 |

- `scripts/build-readme-thumbs.py` regenerates the thumbs; `--check` verifies without writing, suitable as a CI gate. Requires Pillow, fails with a clear message without it. Follows the `build-icons.py` convention: generated files are committed, users of the skill never run it.
- Higher quality on the three inline figures keeps the dot grid and the small caps legible at 838 CSS px.
- `docs/screenshots/*.png` unchanged — same files, same paths.

### Rebased onto current `main`

Rebased onto `b827385` — the consolidated **2.6.0** release (#134), which landed while this was in review and brought the polar type, the render linter, and the doctor/sankey/polar gates.

- **README documentation preserved.** Rather than resolve the conflict by hand, `main`'s README was taken wholesale and the preview rewrite re-applied mechanically over it. Undo that rewrite and the file differs from `main` by exactly **4 lines**, all in the architecture tree — everything else on `main` survives verbatim.
- **All three manifests bumped** — Claude, Codex, Factory — 2.6.0 → **2.6.1** via `scripts/bump-plugin-version.py`.
- **Thumbs regenerated for all 41 screenshots**, including `polar` from this release.

### Addressing the Greptile review

Three findings across the review rounds, all now fixed. The freshness check went through two rewrites because each round was a fair hit.

**P1 — "manifest match skips artifact validation" (latest).** Correct: the manifest pinned the *source* PNG but nothing pinned the thumb itself, so a swapped, corrupted, or hand-edited WebP passed while the README served a broken preview. I had applied only half the `verify-screenshot-freshness.py` convention, which pins both sides.

Each manifest entry now records `thumb_sha256` as well as `source_sha256`, and `--check` decodes every thumb it accepts and confirms its width. Verified against four tampering scenarios:

| scenario | source | manifest | result |
|---|---|---|---|
| thumb swapped for a different image | unchanged | unchanged | **caught** — digest mismatch |
| thumb corrupted in place | unchanged | unchanged | **caught** — digest mismatch |
| thumb truncated | unchanged | unchanged | **caught** — digest mismatch |
| thumb corrupt **and** its digest updated to match | unchanged | consistent | **caught** — fails to decode |

The last row is why the decode step exists: a digest alone cannot tell a valid thumb from a corrupt one that was recorded as-is.

**P1 — "localized edits pass freshness check" (previous round).** The mean-pixel comparison is gone. Digests do not average, so a **single flipped pixel** in `venn.png` — 0.000036% of the image — is reported stale. Still verified.

**P2 — "avoid encoder-dependent freshness checks" (first round).** `--check` no longer runs the encoder at all, so libwebp version and platform cannot affect the result. It is also much faster.

Orphan cleanup still works and never removes the manifest; a thumb that fails to decode is reported rather than silently deleted, and a failure is fatal in build mode too, not only under `--check`.

### Validation gates

Full CONTRIBUTING suite for 2.6.0 re-run green on the rebased branch — **35 gates**, each run individually so no failure could hide behind a `&&` chain. That includes the gates new in this release: `lint-render.py --self-test` and `--all` (141 files rendered under Chromium, 0 findings), `verify-polar.py`, `verify-sankey.py`, `verify-doctor.py` and their adversarial test counterparts, plus `verify-screenshot-freshness.py` and `build-readme-thumbs.py --check`.

`ruff check --target-version py39` is clean on the new script, and it uses no syntax past 3.9, so it stays inside the repo's Python 3.9 compatibility contract.

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples — no examples changed; all `alt` text preserved verbatim

## Two notes for the reviewer

**Same-tab previews.** GitHub auto-wraps un-linked README images in a `target="_blank"` link to their blob page. Explicit `<a>` wrappers are needed here so clicks land on the full-res PNG instead of the 600px WebP — but `/markdown` strips `target`, so previews now open in the **same tab** rather than a new one. Same destination, same hover URL. Happy to drop the wrappers if you'd rather keep new-tab behavior, at the cost of the full-res PNGs no longer being reachable from the README.

**One unreferenced thumb.** `quadrant-consultant.png` is still in `docs/screenshots/` but no longer appears in the README after the catalog refresh, so its 4 KB thumb is generated and unused. The script's contract is "one thumb per PNG"; narrowing it to "only what the README references" would couple the script to README parsing. Say the word if you'd rather I drop the PNG or narrow the contract.

## Related issues

Fixes #77.


iriekarl (approved):

cathrynlavery:
This is the right direction for #77 and the thumbnail check is compelling. Please rebase onto current `main`, preserve all README/profile documentation added since your base, bump both manifests above the new main version, add `Fixes #77`, and rerun the current full CONTRIBUTING suite plus `build-readme-thumbs.py --check`. I’ll re-review the surgically rebased result.

cathrynlavery:

cathrynlavery (changes requested):
The thumbnail approach and rendered output look good, but this head is no longer mergeable against current `main`. Please:

1. Rebase on current `main` and resolve the README/gallery conflicts while preserving the right-sized preview changes.
2. Re-bump all three manifests to one version that is greater than `main` and unique across open PRs.
3. Run the complete current gate set, including the ridgeline and bubble verifier/test pairs that are absent on this branch, plus the full render pass and thumbnail-freshness check.

On exact head `3cee932`, the available render/thumbnail checks passed; the remaining failures are release/base drift rather than a rejection of the image fix itself.

cathrynlavery:
Thank you, @alexwickstrom. The accepted core of this PR landed through #174 with your authorship preserved in commit b8af598. The integrated version covers the current gallery and adds exact-dimension, wiring, adversarial, CI, policy, and documentation checks. This head is fully superseded and conflicts with current main, so I’m closing it rather than asking for duplicate rework.

PR #54, opened 2026-08-14
fix(mermaid): parse quoted participants, create, and the full sequence arrow vocabulary

guanyifang344:
## What does this PR do?

Fixes the shipped `mermaid_extract.py` sequenceDiagram parser, which silently dropped or hard-rejected Mermaid-valid constructs that real user files contain. Every form below was verified against the real Mermaid parser (v11.16.1) before fixing.

### Bugs fixed

1. **Quoted participant/actor names were silently dropped.** `participant "Alice Smith"` (no alias) vanished from the IR entirely; `participant "Alice Smith" as A` and `actor "Bob Builder" as B` lost both their display names and aliases.
2. **`create participant B` / `create actor B` directives were ignored.**
3. **Bidirectional messages hard-failed.** `A<<->>B` and `A<<-->>B` (Mermaid v11.0+) aborted the whole file with `malformed edge` — valid Mermaid rejected outright.
4. **Open arrows were misclassified.** `->` / `-->` are "solid/dotted line **without** arrow" per the Mermaid docs, but were emitted with a filled `arrow` arrowhead.
5. **Declared actor shapes were clobbered.** A message touching a participant declared as `actor` force-overwrote its shape to `lifeline`, because message endpoints were unconditionally auto-added as lifelines.

### What changed

- `participant_re` now accepts quoted names (double- or single-quoted) with or without an `as` alias, plus the `create` prefix. For `participant "Alice Smith" as A` the alias is the id (messages reference it) and the quoted string is the display name, mirroring how the bare form works.
- `message_re` learns the bidirectional `<<->>` / `<<-->>` arrows (retained as `bidirectional` with solid/dashed style) and multi-word endpoints introduced by quoted declarations, while lazy ids keep compact forms like `A-->>B` intact.
- Open arrows `->` / `-->` now emit `arrowhead: none` + `undirected: true`; `-)` / `-x` async/cross handling is unchanged.
- Message endpoints are auto-added only when undeclared, so earlier `actor`/`participant` declarations keep their shape and label.
- `references/import-mermaid.md` documents the supported sequence grammar.

### Verification

- New gate `check_sequence_grammar_forms` in `scripts/verify-mermaid-import.py` covers every form above plus a malformed-bidirectional negative case.
- All 18 repository gates pass locally, including `verify-mermaid-import.py`, `lint-skin.py --all --baseline`, `verify-plugin-package.py`, and `verify-docs-sync.py`.
- `claude plugin validate . --strict` passes.
- Byte-identical output vs the pre-fix extractor on the three shipped fixtures (`sample-flowchart.mmd`, `sample-readme-with-mermaid.md`, `sample-adversarial.mmd`) — no regressions on previously supported grammar.
- Plugin version bumped to 2.4.1 (rebase onto the new 2.4.0 main) in both native manifests (required for every PR).


Vitanakpapa:
guanyifang344:fix/mermaid-sequence-grammar-fidelity

Vitanakpapa:
guanyifang344:fix/mermaid-sequence-grammar-fidelity

guanyifang344:
Rebased onto the current `main` (`09df49d` — named client profiles) and bumped the plugin version to **2.4.1** to stay above the new base. All 19 local gates pass, including the new `test-verify-docs-sync.py` and `verify-docs-sync.py` reference-link checks.

cathrynlavery:
The parser/reference/test changes pass the targeted Mermaid and skin gates locally. I’m sequencing #75 first because both PRs currently claim plugin version 2.4.1. After #75 merges, please rebase onto the new main, bump both manifests above it, and trigger current-head CI; then this should be next in the merge queue.

cathrynlavery:
Maintainer update: rebased this branch onto current main (`7ed2b20a`) and bumped both manifests to `2.4.2` in `cbfbace`. The complete local CONTRIBUTING/CI-equivalent suite and strict Claude plugin validation passed. I’m authorizing current-head CI now; if it remains green, I’ll approve and squash-merge.

cathrynlavery (approved):
Reviewed on updated head cbfbace: full local gate suite and strict plugin validation passed, and current-head CI is green across all required jobs. Approving for squash merge.

PR #57, opened 2026-08-14
fix(lint): don't require exact token-table headings in style-guide.md

AaronRoeF:
### Problem

`allowed_colors()` looks up three headings by exact string and raises `ValueError` when one is missing, so a custom skin can't run the linter at all.

Repro on `main`:

1. Delete the `### Terminal skin` section from `style-guide.md`. It's an opt-in palette — a skin that doesn't use the terminal primitive has no reason to carry it.
2. `python3 scripts/lint-skin.py skills/diagram-design/assets/example-architecture.html`

```
ValueError: missing '### Terminal skin' in .../references/style-guide.md
```

The same rigidity bites from the other direction: a skin that adds its own token tables — a tint ramp, per-surface keylines — gets a wall of `is not in the style-guide palette` findings for values documented on the very next line of the style guide.

### Fix

Scan every Markdown table in `style-guide.md` instead of three named headings. The style guide stays the single source of truth, which is the design system's stated premise — the linter now follows it rather than a hardcoded index of it.

### Verification

`python3 scripts/lint-skin.py --all --baseline` → **102 files, 0 findings**, unchanged from `main`. No loosening for the shipped skin; the crash and the false positives both go away for custom ones.

### Context

Found while skinning the system to a brand palette. Thanks for building this — the progressive-disclosure structure made it genuinely pleasant to work in.


cathrynlavery (changes requested):
The custom-skin goal is valid, but scanning every Markdown table for hex values can silently whitelist unrelated examples or prose tables. Please constrain discovery to declared skin-token tables (including custom token sections), add tests proving custom sections pass while unrelated table hexes remain rejected, then rebase and add the synchronized manifest bump.

cathrynlavery:

AaronRoeF:
Thanks — you were right that the global scan traded one failure mode for a quieter one. Reworked in de8e924:

- **Discovery is now constrained to declared token tables.** A table contributes colors only when its header row names a token-shaped column (`hex`, `token`, `role`, `light`, `dark`, `default`); fenced code is skipped entirely. Custom skins can add their own token sections or drop the opt-in ones, and a stray hex in a prose table or worked example can no longer widen the palette.
- **Tests** in `scripts/test-lint-custom-skin.py`, wired into both CI call sites: an equivalence test proving the shipped guide's palette is identical to the previous three-heading lookup; a custom token section accepted; a guide missing the shipped opt-in sections not crashing; and the rejection cases you asked for — a prose-table hex stays out of the palette *and* a diagram using it still gets a color finding end-to-end, plus fenced-block hexes rejected.
- **Rebased onto current main** (2.5.6) and both manifests bumped to 2.5.7 — `verify-plugin-package.py origin/main` passes. If #59 lands first I'll re-bump.

`--all --baseline`: 108 files, 0 findings, on Python 3.13 and 3.9.

AaronRoeF:
Rebased onto current main (post-2.6.1) and re-bumped all three manifests — including the new `.factory-plugin/` surface — to 2.6.2; `verify-plugin-package.py origin/main` passes. No changes to the reviewed content: the token-table discovery and all 8 tests are identical, `--all --baseline` 141 files / 0 findings.

cathrynlavery (changes requested):
The heading-flexibility goal is good, but both unresolved parser defects still reproduce on exact head `0dd11a7`:

- Ordinary prose containing `Role` can be classified as a token table.
- Fence-delimiter state can desynchronize, causing later content to be parsed in the wrong mode.

Please fix both with focused adversarial fixtures, resolve the review threads, rebase on current `main`, bump all three manifests to a unique increasing version, and run the complete current suite (including the bubble verifier/test pair added after this branch).

cathrynlavery:
Circling back on my Aug 25 review with concrete fixtures, since both parser defects still reproduce on `0dd11a7` — these should drop straight into `scripts/test-lint-custom-skin.py` as the adversarial cases:

**1. Prose tables widen the palette.** Any first table row containing a token word qualifies the whole table, so this prose table:

```
| Role of tool | Example |
|---|---|
| neon look | `#39ff14` |
```

feeds `#39ff14` into the allowed palette via `iter_token_table_rows()`. Classification needs to be stricter than a word match anywhere in the header — e.g. match column names exactly after splitting on `|`, or require the table to actually declare a hex-bearing token column.

**2. Fence state desynchronizes.** The unconditional `in_fence = not in_fence` fires on any line starting with three backticks *or* `~~~`, so a backtick fence containing a `~~~` line — or a four-backtick fence containing a three-backtick line — exits the fence early, and a token-shaped table inside the example then contributes its hexes (I reproduced with `#abc123` and `#def456`). Track the opening delimiter's character and length and only close on a matching or longer marker of the same character.

Housekeeping: main is at 2.6.11 now, and #57 and #59 both currently carry 2.6.5, so they also collide with each other — whichever you rebase second needs its own higher version. The manifest bumps are the only real merge conflict; `scripts/lint-skin.py` and `ci.yml` merge cleanly onto current main.

Everything else holds up in a clean worktree at your head: `test-lint-custom-skin.py`, `test-lint-a11y.py`, and `lint-skin.py --all --baseline` (144 files, 0 findings) all pass. Once the two fixtures above go red-then-green, the threads are resolved, and the branch is rebased with a fresh bump, this looks ready for a final pass.

cathrynlavery (changes requested):
The heading-flexibility goal remains valid, but exact head is unsafe: prose headers containing token words can whitelist hex values, and fence tracking toggles on mismatched delimiter characters/lengths. Tighten table classification to exact column semantics, track opening fence character/length, and add red-then-green fixtures. Rebase onto main, drop manifest bumps under ADR 0009, and update to the current gate set.

cathrynlavery (changes requested):
I really like the goal here: custom style guides should be allowed to use honest section names without weakening validation. Rechecked exact head `f1a205b` against current main: both unresolved parser defects still block this change. Prose tables can widen the accepted palette, and fenced delimiter text can desynchronize section parsing. Make heading flexibility operate on a quote/fence-aware Markdown structure with adversarial fixtures for both cases. Then rebase onto current main, drop manifest/version edits, resolve both threads, register against the complete current gate set, and rerun all 58 commands; this head passes only 53. I appreciate the care you’ve put into this.

cathrynlavery:
The custom-heading goal is valid, but this head still has two unresolved fail-open parser paths and is far behind the current gate set. I’m closing the branch to keep the queue actionable. A fresh PR is welcome using the existing prose-table and mixed-fence fixtures, with quote/fence-aware parsing and the current full suite.

PR #58, opened 2026-08-14
fix(lint): restore Python 3.9 compatibility

AaronRoeF:
### Problem

`lint-skin.py` uses PEP 604 annotations (`str | None`) in dataclass fields. Those are evaluated at class-creation time, so the script raises on Python 3.9 before doing any work:

```
$ /usr/bin/python3 scripts/lint-skin.py skills/diagram-design/assets/example-architecture.html
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

3.9 is still what macOS ships as `/usr/bin/python3`, so a contributor following CONTRIBUTING's `python3 scripts/lint-skin.py <file>` hits this before running the linter once — and the error doesn't hint at the cause.

### Fix

`from __future__ import annotations`. One line, defers annotation evaluation, keeps the modern syntax everywhere.

### Verification

- Python 3.9.6 (`/usr/bin/python3`), single file → 0 findings, runs clean
- Python 3.13, `--all --baseline` → 102 files, 0 findings

Happy to switch to documenting a 3.10+ floor in CONTRIBUTING instead if you'd rather set the requirement than widen support — your call.


cathrynlavery (changes requested):
The `from __future__ import annotations` fix works locally on Python 3.9.6. Before merge, please rebase onto current main, bump both manifests in sync, and add a Python 3.9 compatibility job (a focused lint/a11y smoke job is sufficient) so the compatibility claim remains enforced rather than relying on a one-time local check.

cathrynlavery:

cathrynlavery (approved):
Current-head Greptile review is 5/5, the Python 3.9 compatibility job is green, the full seven-job CI matrix passed, and the complete local gate suite passed on commit 7259cbd.

PR #59, opened 2026-08-14
fix(export): don't require Chromium specifically for PNG export

AaronRoeF:
### Problem

`references/export.md` hardcodes `p.chromium.launch()`, and the detection step tells the user to run `playwright install chromium`. Someone who already has WebKit or Firefox installed is told to download a second engine before they can export a PNG.

Nothing in the export path is engine-specific — it opens a local HTML file and screenshots one element.

### Fix

Try the engines in turn and use whichever is present; only ask the user to install one if none is. The install hint now names WebKit with a note that any engine works, rather than implying Chromium is required.

### Verification

Verified on **WebKit only** — the engine on this machine. Correct element framing and transparent background at `device_scale_factor` 2 and 3, on both a shipped example and a custom-skinned diagram. I did not diff against Chromium output, since there's no Chromium here to compare against.

### Note

The SVG-export step just above this injects a hardcoded `@import` of Instrument Serif / Geist / Geist Mono. Under a custom skin that exports type which doesn't match the diagram it came from. I left it alone here to keep this PR to one concern, but happy to send a follow-up deriving that font list from `style-guide.md` if you'd like it.


cathrynlavery (changes requested):
The engine fallback direction is useful, but changes are required. Please rebase from the stale 2.3.5 base and bump both manifests above current main. Narrow the broad `except Exception`: dependency, permission, browser-crash, and configuration errors must not all become the misleading “no engine installed” message. Either catch only the expected missing-executable condition or preserve per-engine diagnostics, and add a focused test for fallback plus error reporting.

cathrynlavery:

AaronRoeF:
All four asks addressed in 2112019:

- **Narrow failure filter.** The rasterize step is now a real script, `skills/diagram-design/scripts/export_png.py` (inside the skill, next to the extractors, so plugin installs ship it). Only Playwright's `Executable doesn't exist` launch failure advances the WebKit → Chromium → Firefox fallback; a crash, permissions, or configuration error propagates with its real message. Total-miss exits with the per-engine list plus the install hint.
- **Focused test.** `scripts/test-export-png-fallback.py` exercises the fallback with fake launchers (no Playwright needed in CI): falls back past a missing engine, stops at the first installed one, propagates a `PermissionError` unmodified, lists every engine on total failure, and the filter matches only Playwright's missing-executable wording. Wired into both CI call sites; green on 3.13 and 3.9.
- **README fixed too** — the one-time-setup line still said `playwright install chromium` (good catch by Greptile); now any one engine.
- **Rebased onto 2.5.6**, both manifests bumped to 2.5.7, `verify-plugin-package.py origin/main` passes. Whichever of #57/#59 merges second gets a trivial re-bump from me.

One honest scope note: I haven't run a live rasterize under pip-Playwright on this machine — the fallback logic is covered by the fakes, and the Playwright API calls are unchanged from the previous snippet apart from the launch loop.

AaronRoeF:
Same maintenance: rebased onto current main, all three manifests (incl. new `.factory-plugin/`) to 2.6.2, gate passes. Reviewed content unchanged — fallback script, tests, and README fix identical.

cathrynlavery (changes requested):
Two portability gaps remain on exact head `2e24727`:

1. The docs still run `python scripts/export_png.py`, which resolves against the caller's working directory and fails for an installed skill outside the repo. Resolve the script from the installed skill directory.
2. Doctor still checks Chromium specifically even though the export path now accepts WebKit or Firefox. Update doctor, docs, and tests so any supported installed engine satisfies the diagnostic.

Please also reopen/resolve the review threads against the code that actually lands, rebase on current `main`, choose a unique synchronized version, and run the complete current gate set.

cathrynlavery (changes requested):
The Chromium fallback works, but this head still needs a current-main rebase with manifest edits dropped. Resolve the helper from the installed skill, make doctor/docs accept any supported installed engine, size output deterministically from viewBox, qualify transparency, append/assert static motion, and use an end-to-end bounded font/network path covering #176. Add installed-skill and real rendered-export tests, then run current gates.

cathrynlavery (changes requested):
Thanks for pushing the PNG path toward genuine cross-engine portability. Exact head `f21dca7` remains behind current main and passes 53 of 58 current commands. The engine fallback alone is useful, but the shipped path still needs to resolve the helper from the installed skill, make doctor/docs accept any supported installed engine, derive deterministic size from the viewBox, qualify transparent-background behavior, force/assert static motion, and bound font/network waits with real rendered-export coverage. Rebase onto current main, remove manifest bumps, add the installed-skill and rendered fixtures, and run the complete policy suite. I appreciate the care you’ve put into this.

cathrynlavery:
The cross-engine goal is still useful, but this branch has been overtaken by the current export work and now mixes several unresolved contracts from later reviews. I’m closing it rather than asking for another high-conflict rebase. After #183 and #235 settle the timeout and standalone-export paths, a fresh narrow multi-engine fallback PR would be welcome.

Issue #65, opened 2026-08-15
Type references hardcode the default skin; zone label opacity breaks on a custom paper


AntonioIbarraOrtiz:
## Summary

The type references hardcode the **default skin's** hex values and font families in their SVG snippets, which contradicts the contract `SKILL.md` §5 states:

> When specs below or in type references mention "ink", "accent", "muted", etc., look up the current hex value in `style-guide.md`.

Measured on 2.4.0, excluding `style-guide.md`, `onboarding.md`, `profiles.md` and `primitive-terminal.md` (the terminal skin is intentionally fixed):

**23 files, 207 occurrences** of `#2d3142`, `#f5f5f5`, `#eb6c36`, `#4f5d75`, `#7a8399`, `#bfc0c0`, `#2e5aa8`, `rgba(45,49,66,…)`, `rgba(235,108,54,…)`, `Instrument Serif`, `Geist`.

| File | Occurrences |
| --- | --- |
| `type-process.md` | 30 |
| `type-high-level.md` | 27 |
| `type-dp-integration.md` | 25 |
| `type-sequence.md` | 15 |
| `type-quadrant.md` | 9 |
| `type-gantt.md` | 9 |
| `type-bar.md` | 8 |
| `type-scatter.md` | 8 |
| `output-spec.md` | 8 |
| `type-layers.md` | 5 |
| `type-radar.md` | 5 |
| `type-venn.md` | 4 |
| …plus 11 more | 1–3 each |

Mostly this is a trip-hazard rather than a hard break — an agent that follows §5 substitutes correctly. But **one case is an actual defect**, because the hardcoded value encodes an assumption about the default paper.

## The concrete defect: zone label opacity

`type-architecture.md` § Zone grouping specifies the zone eyebrow as:

```svg
<text … fill="rgba(45,49,66,0.40)" font-size="7"
      font-family="'Geist Mono', monospace" …>LAYER</text>
```

`ink @ 0.40` was evidently tuned against the default `paper` (`#f5f5f5`). Onboard a lighter, warmer paper and it degrades badly. With `paper #faf8f6` / `ink #2f3235`:

| Alpha | Composite | Contrast on paper |
| --- | --- | --- |
| **0.40 (shipped)** | **`#a9a9a9`** | **2.22 : 1** |
| 0.50 | `#949596` | 2.83 : 1 |
| 0.65 | `#767779` | 4.23 : 1 |

[...]

PR #66, opened 2026-08-15
fix: Reference semantic roles instead of the default skin's hex values

AntonioIbarraOrtiz:
## What does this PR do?

Converts the type references from the default skin's hardcoded hex values and font families to the semantic-role placeholders the rest of the project already promises — plus three defects that hardcoding caused, the most serious of which makes a generated diagram fail this repo's own linter.

This is the follow-up to #65. I opened that as an issue first and offered to do the work; rather than leave you a 25-file audit to act on, I've done the conversion so accepting it is a review rather than a project. **Happy to split, reduce, or revert any part of this** — see *Open question* and *Deliberately left alone* below for the two places I made a judgement call that's really yours.

### Why

`style-guide.md` opens with:

> Every diagram draws from this — **not from hex values inlined in other reference files**.

and `SKILL.md` §5 tells the agent to look up the current value of a named role. The type references predate that contract and inline the default skin directly, so a reader copying a snippet verbatim emits off-skin output.

Two things convinced me this was an unfinished migration rather than a deliberate choice:

- `type-it-state.md` and `type-data-flow.md` — the newest references — **already** use `{ink}`, `{ink @ 0.40}`, `{muted}`, `{node-name}`, `{sublabel}`. I adopted their notation exactly rather than inventing one.
- `type-architecture.md`'s zone snippet is **half** converted: geometry and `{paper}` are placeholders in the same four lines where the colors and fonts are literal.

### Scale

Measured against `main` with the full default-skin pattern set: **25 files, 265 occurrences.** (Issue #65 says 23/207 — that was measured against a working copy where I'd already fixed two files, with a narrower pattern. This is the corrected figure.)

### The three real defects

**1. Zone label opacity — `fix(architecture)`.** The zone eyebrow was `ink @ 0.40`, an opacity tuned against the default `#f5f5f5` paper. Being an opacity rather than a role, it doesn't track the skin. With `paper #faf8f6` / `ink #2f3235`:

| Alpha | Composite | Contrast on paper |
| --- | --- | --- |
| **0.40 (as shipped)** | **`#a9a9a9`** | **2.22 : 1** |
| 0.50 | `#949596` | 2.83 : 1 |
| 0.65 | `#767779` | 4.23 : 1 |

At 7px uppercase tracked that's illegible, and fainter than the skin's own `soft` token (`#71777d`, 4.27:1 there). `style-guide.md` already defines `soft` as *"Sublabels, boundary labels"* — a zone eyebrow **is** a boundary label, so the role existed all along and the magic number was the defect, not its value. Found while generating a diagram on a custom brand skin, where this was the only element that read as broken.

**2. Off-palette chevron fills — `docs(high-level)`.** `type-high-level.md` documents `#3d4460` / `#4a5270`, which aren't in the palette. Verified rather than assumed:

```
$ python scripts/lint-skin.py <probe file using the documented chevron fill>
color: #3d4460 is not in the style-guide palette
```

So an agent following that reference literally **produces a diagram that fails CI** — which is, I think, why all six shipped high-level examples sit in `scripts/lint-skin-baseline.txt`.

**3. Export font import — `fix(export)`.** Step 3 of the SVG export procedure hardcoded the default Google Fonts URL. On a re-skinned project that injects the wrong families into the `.svg`: text set in the brand's faces, `@import` fetching Instrument Serif / Geist / Geist Mono, every label silently falling back. Nothing warns. The source HTML already carries the right stack in its `<link href>`, so the fix derives the `@import` from it.

### Side effect: the dark-mode tables shrink

Most rows restated by hand what role inversion already does — `ink @ 0.12` resolves to dark ink on light paper and light ink on dark paper on its own. Only rows where the *opacity* genuinely differs between modes survive, each with a one-line note on why. `type-process.md` §6 goes from 16 rows to 6, `type-dp-integration.md` from 13 to 5.

### Open question — the phase-band palette (your call)

The chevron banner needs a dark band with `paper`-colored labels in **both** modes, so it genuinely can't be built from `ink` — in dark mode `ink` *is* the light color. I did **not** silently restyle it. Instead I named the three steps as a `band-1` / `band-2` / `band-3` palette in a new §6.1 and stated plainly that they aren't semantic roles, don't follow the active skin, and are rejected by the linter, with instructions for deriving three analogous steps from a new brand's darkest neutral.

Whether they should instead become real tokens in `style-guide.md` (which would let the six baselined examples come off the baseline) is a design decision, so I left it to you. Say the word and I'll do it either way.

### Deliberately left alone

Because the literal is the point, not an oversight:

- the concern palette (rust-red / slate-blue / olive-green / warm-brown) and the data-type chip colors — separate semantic axes from the skin
- user-supplied `color:` overrides in the input contracts
- `onboarding.md` naming the defaults it checks for, `primitive-terminal.md` instructing you to *drop* those families, and `import-drawio.md` mapping a draw.io source color, which is untrusted input rather than our skin
- `style-guide.md` itself — hex there is correct by definition

**No rendering change** except the two fixes above: every other substitution resolves to the value it replaced under the default skin.

## Validation gates

Ran the full suite from CONTRIBUTING.md, all passing. I also ran it on unmodified `main` first so I could tell a pre-existing failure from one I introduced — there were none either way.

- [x] `git diff --exit-code` green after `python3 scripts/build-icons.py` (no icon changes; ran it anyway to confirm nothing went stale)
- [x] Docs updated in the same PR where behavior changed — this PR *is* the docs

Also green: `test-plugin-package.py`, `verify-plugin-package.py origin/main`, `verify-semantic-motion.py` (both modes), `verify-motion.py --shipped`, `test-verify-motion.py`, `verify-docs-sync.py`, `test-verify-docs-sync.py`, `test-self-check.py`, `verify-geometry.py --all`, `test-verify-geometry.py`, and `claude plugin validate . --strict`.

Version bumped to **2.4.1** via `scripts/bump-plugin-version.py`, so both manifests stay in sync.

One environment note: I ran these as `python` rather than `python3` — on Windows `python3` is the Microsoft Store shim. Python 3.14.5.

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples — no examples changed; the only SVG I built was a throwaway linter probe, not committed

## Related issues

Addresses #65.

Leaving the phase-band token decision open rather than closing #65 outright — that thread is the better place to settle it if you'd rather the bands become real tokens.


cathrynlavery (changes requested):
The semantic-role migration is strong and its existing gates pass, but there is a blocking dark-mode regression in `type-high-level.md`: chevron labels remain `paper` while dark mode uses `band-2`/`band-3`, producing only about 1.35:1–1.68:1 contrast. Please use a label role/token that preserves at least 4.5:1 in both modes. Also resolve the phase-band policy by adding formal tokens/linter support or documenting and testing a deliberate type-scoped exception. Rebase and bump above main after #75 lands, and add `Fixes #65` when ready.

cathrynlavery:

cathrynlavery (changes requested):
The semantic-role direction is right, but this branch still has blocking behavior:

- High-level dark phase bands keep paper-colored labels at roughly 1.35–1.68:1 contrast. Formalize phase-band tokens (and any intentional linter exception) so the dark treatment stays readable.
- The unresolved P1 is valid: already escaped font URLs are double-escaped (`&amp;` becomes `&amp;amp;`). Preserve URL query strings exactly once.
- Add `Fixes #65` so the issue closes only with the implementation.

Then rebase and resolve conflicts, bump all three manifests to a unique current version, resolve the threads, and run the full current gate suite; exact head `358905b` contains only 22 of the 41 current commands.

cathrynlavery (changes requested):
The semantic-role migration remains right, but rebase and drop manifest edits under ADR 0009; preserve current CJK guidance. Resolve phase-band labels that reproduce at 1.346:1/1.677:1, define an AA-safe small boundary-label contract, and make Google Fonts href handling decode once then XML-escape once for both bare & and &amp; inputs. Restrict to the intended endpoint, add regression tests, resolve the export thread, and run all current gates.

cathrynlavery:
Thank you for carrying this broad semantic-role cleanup through several rounds of review. Exact head `b5569a7` is current-base, conflict-free, and passes all 58 local gates. The current code appears to address the phase-band contrast, small boundary-label, CJK-preservation, and decode-once/XML-escape-once concerns. One policy blocker is still visible: the outdated export thread remains unresolved. Please resolve it and confirm whether this PR intentionally leaves #65 open (`Addresses #65`) or should close it. After that, the exact first-time workflow run can be considered separately and the resulting current-head CI reviewed. I appreciate the care you’ve put into this.

PR #69, opened 2026-08-15
Add adversarial tests for verify-semantic-motion.py and verify-sequence-oauth.py

alvinhui:
## Why

Every other verifier under `scripts/` has a matching `test-verify-*.py` (verify-docs-sync, verify-geometry, verify-motion, verify-plugin-package), and CONTRIBUTING.md lists both `verify-semantic-motion.py` and `verify-sequence-oauth.py` as required PR gates — but neither one had a dedicated test file exercising its own pass/fail behavior. This PR closes that gap by following the exact pattern already established by the other test files.

## What's included

- **`scripts/test-verify-semantic-motion.py`** — asserts the shipped `SKILL.md` / `semantic-patterns.md` / `animation.md` and the policy-trace animated example pass `verify_markdown()` / `verify_example()`, plus three adversarial cases:
  - a missing semantic-pattern router link (`semantic-patterns.md` reference removed)
  - a missing pattern name (one of the seven named patterns dropped)
  - a duplicate HTML/SVG `id` injected into the animated example
- **`scripts/test-verify-sequence-oauth.py`** — CLI smoke-tests the shipped repo (expects `ALL GATES PASSED`), plus an adversarial case that removes the dashed-return marker from the OAuth 401 response in a temporary copy of the repo tree. Note: `verify-sequence-oauth.py` resolves all paths from its own `__file__`, so the adversarial run has to execute the *copied* script against the *copied* tree rather than just changing `cwd`.
- Both new tests are wired into `CONTRIBUTING.md`'s gate table and the combined validation command, matching the existing `test-verify-*.py` entries.

I intentionally left `.github/workflows/ci.yml` untouched in this PR (my token doesn't have the `workflow` scope to push changes there) — happy to add the CI wiring in a follow-up if that's useful, or a maintainer can add the two `python scripts/test-verify-semantic-motion.py` / `python scripts/test-verify-sequence-oauth.py` steps alongside the existing `verify-sequence-oauth.py` / `verify-semantic-motion.py` steps.

## Testing

Both new test files pass locally against the current `main`:

```
$ python3 scripts/test-verify-semantic-motion.py
OK: shipped SKILL.md / semantic-patterns.md / animation.md pass verify_markdown
OK: shipped policy-trace example passes verify_example
OK: missing semantic-pattern router link is rejected
OK: missing semantic-pattern name is rejected
OK: duplicate HTML/SVG id in the animated example is rejected
All semantic-motion verifier tests passed.

$ python3 scripts/test-verify-sequence-oauth.py
OK: shipped repo passes verify-sequence-oauth.py
OK: removing the dashed 401 return marker is rejected
All sequence-oauth verifier tests passed.
```

Also re-ran the existing adjacent gates (`verify-semantic-motion.py --markdown-only`/`--example-only`, `verify-sequence-oauth.py`, `verify-docs-sync.py`, `test-verify-docs-sync.py`) to confirm nothing else regressed.


cathrynlavery (changes requested):
Thanks — the new adversarial tests are useful and pass locally. Changes requested before merge: (1) add both new test commands to `.github/workflows/ci.yml`, since CONTRIBUTING now declares them as gates; (2) bump both plugin manifests in sync above current main. As a non-blocker, consider avoiding a full repository copy for the single OAuth mutation fixture.

cathrynlavery:

alvinhui:
Thanks for the review! Pushed a follow-up commit addressing all three points:

1. Wired `test-verify-semantic-motion.py` and `test-verify-sequence-oauth.py` into `.github/workflows/ci.yml` as their own steps (with matching step-summary rows), so CONTRIBUTING's declared gates actually run in CI.
2. Bumped both `.claude-plugin/plugin.json` and `.codex-plugin/plugin.json` in sync to `2.5.7` via `scripts/bump-plugin-version.py`; `python3 scripts/verify-plugin-package.py origin/main` now passes.
3. Non-blocker: simplified `test-verify-sequence-oauth.py`'s adversarial fixture to copy only the files `verify-sequence-oauth.py`'s `main()` reads before the dashed-return check (`type-sequence.md` + the cold-cache/oauth example trios) instead of `shutil.copytree` of the whole repo.

Ran the full local test suite (`scripts/test-*.py`) plus `verify-plugin-package.py origin/main` — all green.

philyuchkoff (commented):
# Review: adversarial tests for verify-semantic-motion and verify-sequence-oauth | PR #69

@alvinhui 

Clean closure of a gap — these two verifiers were the only ones without their own test files, and now they have them. Both follow the house pattern (pass-on-shipped, fail-on-adversarial, mutation on temporary files).

**test-verify-semantic-motion.py:**

Positive smoke loads the real module via importlib.util and calls its entrypoints. Three negatives: missing router link, missing pattern name, duplicate id in animated example. All use tempfile for isolation. The importlib.util approach exposes internals (SKILL, EXAMPLE) for mutation — but that is the same testability tradeoff already in this repo (test-verify-docs-sync.py uses verify.GALLERY = ...). Fine.

**test-verify-sequence-oauth.py:**

Smoke test uses subprocess.run to invoke the real verifier — same pattern as test-verify-motion.py. The adversarial case creates a mirror temp directory because verify-sequence-oauth.py resolves ROOT from its own __file__. This is the correct approach — the mutation happens on a copy of the needed tree, not on originals.

**One concern:** The PR body mentions a version bump to 2.6.2, but the diff shows NO version bump in plugin manifests. Every PR touching docs/tests requires bump-plugin-version.py per CONTRIBUTING. The PR is mergeable and checks are green, so Greptile may have passed it. Please confirm the bump is deliberate or add it.

LGTM on code regardless.

alvinhui:
@philyuchkoff good catch, thanks. The earlier 2.5.7 bump was real at the time, but it got absorbed by later `Merge branch 'main' into add-verifier-tests` merges (main itself moved past it up to 2.6.5), so the branch had quietly settled back to zero version delta vs. main — not an intentional 2.6.2, that number doesn't appear anywhere in the PR body or diff.

Just pushed a follow-up commit that re-runs `scripts/bump-plugin-version.py` against current `main`, bumping `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, and `.factory-plugin/plugin.json` to `2.6.6`. Confirmed `python3 scripts/verify-plugin-package.py upstream/main` passes, and re-ran the full `scripts/test-*.py` suite locally — all green.

PR #82, opened 2026-08-18
docs(readme): fix slash-command names to match shipped commands

FakeRocket543:
The README documents `/diagram-design:import` and `/diagram-design:export`, but the shipped command files are `commands/import-drawio.md` and `commands/export-diagram.md`, which register as `/diagram-design:import-drawio` and `/diagram-design:export-diagram`. The import-mermaid and Pi examples already used the correct names, so a user's first attempt at the draw.io import or the export command fails with unknown-command.

This PR renames the six occurrences (README.md:246-248, 297-299) to match the shipped commands. No behavior change.


cathrynlavery (approved):
Current-head Greptile review is 5/5, the documentation matches the shipped command filenames, both manifests are synchronized at 2.4.5, and the complete local and current-head CI suites are green.

PR #84, opened 2026-08-18
fix(verify-geometry): recognize label masks up to 200px wide

ggplab:
## What does this PR do?

`verify-geometry.py` recognizes a label mask only when it is 20–120px wide, and ADR 0005 justifies the window with "the heuristics … match the shipped templates". That premise no longer holds: `example-sequence-oauth.html` ships 128px mono plates (`GET /RESOURCE + BEARER` at line 112, `POST /TOKEN · REFRESH` at line 148, plus the `-dark`/`-full` variants — 6 files total), and CJK labels produce wider plates still at the same glyph count, since a Hangul/Han glyph advances ~1em where Latin averages ~0.55em. Every mask past 120px is silently skipped today, so a clipped wide label ships with the gate green. (These plates also run past the §6 "≤14 characters" guidance — at 22 and 21 characters — so long plates are a shipped reality, not a rule violation this PR introduces.)

This widens `MASK_MAX_W` to 200 and keeps the height cap at 14: raising it to 18 was measured to misclassify ~80 shipped rects (zone eyebrows like it-state's 92×16, nested container header bars at 188×16, ER row stripes) as masks, which would produce nonsense "move the label onto a free segment of its connector" findings. Mask classification across the 104 shipped assets goes 412 → 427 rects; findings stay 0 before and after.

Docstring and ADR 0005's Consequences are updated in the same PR, per the ADR's own instruction: "widen them in the checker, and add the case to the adversarial tests."

Adversarial coverage added:
- a 128×12 plate clipped by a later node **must** be reported — fails against the current constant (measured),
- the same plate over an earlier node must not be,
- a 188×16 header bar overlapping a later node must stay unreported — locking the height-cap decision,
- `example-sequence-oauth.html` joins the shipped-file checks (its newly-visible masks are clean).

## Validation gates

- [x] `python3 scripts/test-verify-geometry.py`
- [x] `python3 scripts/verify-geometry.py --all` (104 files, 0 findings)
- [x] `python3 scripts/test-lint-a11y.py` / `python3 scripts/lint-skin.py --all --baseline` (untouched by this change)
- [x] `python3 scripts/test-plugin-package.py` && `python3 scripts/verify-plugin-package.py <base sha>` && `claude plugin validate . --strict`
- [x] Docs updated in the same PR where behavior changed (docstring + ADR 0005)

## Checklist

- [x] Generated and source files are consistent
- [x] Accessible SVG contract satisfied for new/changed examples (no examples changed)

## Related issues

None open; this follows ADR 0005's standing instruction for revisiting the thresholds. Note: bumps the plugin version to 2.4.4 like every PR — if another PR lands on `main` first, I'll rebase and re-bump.


cathrynlavery (approved):
Current-head Greptile review is 5/5, the wider mask heuristic is bounded by unchanged height constraints and adversarial tests, both manifests are synchronized at 2.4.7, and the complete local and current-head CI suites are green.

PR #85, opened 2026-08-18
fix(mermaid): parse compact labeled links (B--yes-->C) without phantom nodes

### LINKED ISSUE #80: mermaid_extract mis-parses compact edge-label syntax (B--yes-->C): phantom nodes, labels dropped
## Bug: `mermaid_extract.py` mis-parses compact edge-label syntax (`B--yes-->C`) — phantom nodes, labels dropped

**Version:** v2.4.3 (`e48df3f`) — also present in 2.3.2.

### Repro

```bash
printf "flowchart LR\n A[Start]-->B{Gate}\n B--yes-->C[Done]\n B--no-->A\n" > t.mmd
python3 skills/diagram-design/scripts/mermaid_extract.py t.mmd
```

`B--yes-->C` is standard Mermaid (edge B→C labeled "yes"), but the extractor reports **5 nodes / 3 edges, 0 labeled**: it materializes `B--yes` and `B--no` as phantom nodes (they show up in hubs/entry-points), and every edge label is lost. The spaced form parses correctly:

```bash
printf "flowchart LR\n  A[Start] -- yes --> B{Gate}\n  A -- no --> C[Stop]\n" | python3 skills/diagram-design/scripts/mermaid_extract.py /dev/stdin
# → 3 nodes / 2 edges, 2 labeled ✔
```

### Impact
Corrupted IR for a very common syntax: node/edge budgets miscounted, focal/hub candidates and entry points polluted, all inline labels dropped — so downstream redraws lose gate semantics.

### Suggested fix
In the edge-operator pass, recognize the compact label forms (`--text-->`, `-.text.->`, `==text==>`, and `--text-->` with `<-->` variants) as a labeled edge rather than treating the label token as node text. Happy to submit a PR with tests if the approach sounds right.

ggplab:
## What does this PR do?

Fixes #80. The labeled-link pattern in `mermaid_extract.py` required whitespace around the label, so the compact form Mermaid also accepts — `B--yes-->C` — fell through to the plain-operator pass: the extractor materialized `B--yes` / `B--no` as phantom nodes (surfacing in hubs and entry points), dropped every label, and the thick compact form `D==crit==>E` aborted the whole import with `malformed edge` (reproduced against `e48df3f`).

The fix accepts a whitespace-free label between the operator halves as a second alternative in the same regex. Two guards keep the near-misses out:

- the compact label may not **start** with an operator character (`-`, `=`, `.`), which keeps `A----->B` an unlabeled long arrow;
- the compact label may not **contain** whitespace, which keeps chained links like `A --o B --> C` two separate links instead of one `A -"o B"-> C`.

The spaced form is matched first and its behavior is untouched.

Adversarial coverage in `verify-mermaid-import.py`: the issue's exact repro (phantom nodes, dropped labels), compact dotted/thick/undirected labels with their styles retained, and the two near-miss forms above asserted to stay unlabeled. Run against the previous extractor, the new block fails immediately (`extractor exited 2 … malformed edge`, measured). `import-mermaid.md` documents the compact form in the same PR.

## Validation gates

- [x] `python3 scripts/verify-docs-sync.py`
- [x] `python3 scripts/test-lint-a11y.py` / `python3 scripts/lint-skin.py --all --baseline` (untouched by this change)
- [x] `python3 scripts/verify-drawio-import.py` (untouched by this change)
- [x] `python3 scripts/test-plugin-package.py` && `python3 scripts/verify-plugin-package.py <base sha>` && `claude plugin validate . --strict`
- [x] Docs updated in the same PR where behavior changed (`import-mermaid.md`)

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples (no examples changed)

## Related issues

Closes #80. Note: bumps the plugin version to 2.4.4 like every PR — if another PR lands on `main` first, I'll rebase and re-bump.


cathrynlavery (approved):
Current-head Greptile review is 5/5, the compact Mermaid edge-label regression test demonstrably fails with the old extractor, the complete local and current-head CI suites are green, and both manifests are synchronized at 2.4.8.

PR #83, opened 2026-08-18
fix(lint): allow Korean fallback stacks alongside the documented Japanese ones

ggplab:
## What does this PR do?

`output-spec.md` §4 promises CJK coverage for "Japanese, Chinese, or Korean" labels, but the only prescribed stack is Japanese-only — Hiragino Sans and Yu Gothic carry no Hangul glyphs — and `lint-skin.py` rejects the natural Korean analog outright:

```
font-family: unsupported font family: Apple SD Gothic Neo, Noto Sans KR, Malgun Gothic
```

This documents the Korean stack next to the Japanese one (`'Geist', 'Apple SD Gothic Neo', 'Noto Sans KR', 'Malgun Gothic', sans-serif`, mono: `'Geist Mono', 'Noto Sans Mono CJK KR', monospace`) and allows its four families in the linter — the same shape as #75 did for the documented Japanese stack. Worth noting: the shipped `cjk-mono-stack` fixture from #75 renders *Korean* text (`포트 443`) through the JP-only stack, which is exactly the gap this closes.

Test coverage mirrors what made #75 land: positive cases for the SVG attribute, mono, and CSS-block forms, plus a sibling-family rejection (`Malgun Gothic Semilight`) proving no prefix matching sneaks in. The three new `require_pass` cases fail against the current allowlist (measured before applying the fix).

## Validation gates

- [x] `python3 scripts/test-lint-a11y.py`
- [x] `python3 scripts/lint-skin.py --all --baseline`
- [x] `python3 scripts/verify-sequence-oauth.py` (untouched by this change)
- [x] `python3 scripts/verify-drawio-import.py` (untouched by this change)
- [x] `python3 scripts/verify-mermaid-import.py` (untouched by this change)
- [x] `python3 scripts/test-plugin-package.py` && `python3 scripts/verify-plugin-package.py <base sha>` && `claude plugin validate . --strict`
- [x] Docs updated in the same PR where behavior changed (`output-spec.md` §4)

## Checklist

- [x] No new entries added to `scripts/lint-skin-baseline.txt`
- [x] Generated and source files are consistent
- [x] Accessible SVG contract satisfied for new/changed examples (no examples changed)
- [x] Code of Conduct respected

## Related issues

Follow-up to #74 / #75, which fixed this for the Japanese families. Note: this bumps the plugin version to 2.4.4 like every PR — if another PR lands on `main` first, I'll rebase and re-bump.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #86, opened 2026-08-18
feat(i18n): add Korean label rules and Noto Sans KR

etaseoul-jason:
## What does this PR do?

Gives Korean labels a font fallback the skill will actually reach for, and puts the rule where a routine request can see it. Today the CJK guidance lives only in `references/output-spec.md`, which loads on the import paths — a plain "draw me a flowchart" request never reads it, so Korean text renders with no fallback stack. The fallback examples that are there are Japanese-only (`Hiragino Sans`, `Noto Sans JP`, `Yu Gothic`), so there is no Korean stack to reach for even on the import path.

Changes:

- A `### Korean labels` section in `references/style-guide.md`, plus a one-paragraph summary in the SKILL.md §5 typography block — both always-loaded surfaces. `output-spec.md` now points at it instead of carrying a second copy.
- `Noto Sans KR` added to the Google Fonts link in SKILL.md and the four templates that carry one, so a Korean diagram renders the same on macOS, Windows, and a reviewer's browser rather than depending on locally installed fonts.
- `ALLOWED_FONTS` in `scripts/lint-skin.py` gains `noto sans kr`, `apple sd gothic neo`, `malgun gothic`.
- The width budget uses measured advances instead of a rule of thumb. In Noto Sans KR at `node-name` size (12px / 600) a Hangul syllable advances **11px**, a space **3px**, and a typical Geist Latin glyph **6.5px**, so a label is `11 * syllables + 6.5 * latin + 3 * spaces`, rounded up to the next multiple of 4 with padding. Measured in Chromium via `getBBox()`: predicted 47px for `상품 등록`, measured 47.0px. Sizing a Korean label off Latin character count overflows the box.
- Korean arrow labels, eyebrows, and legend items drop the 7-8px Geist Mono all-caps treatment. There is no mono Hangul face in the stack and Hangul is illegible at that size, so those slots become 12px sans at weight 500 with the mask rect sized from the formula. Latin labels in the same diagram keep the mono treatment.

`assets/template-terminal.html` is deliberately untouched. It is a mono-only skin by design, and adding a proportional Korean face would break the register it exists for; Korean in the terminal variant needs its own decision.

No new visual type, no change to the connector rules, no behavior change to the extractors. SKILL.md is 38,182 bytes against the 40,000 cap in ADR 0004.

## Validation gates

- [x] `git diff --exit-code` green after `python3 scripts/build-icons.py` (if icons changed) — n/a, icons untouched

Also run green: `test-plugin-package.py`, `verify-plugin-package.py origin/main`, `claude plugin validate . --strict`, `verify-semantic-motion.py` (both `--markdown-only` and `--example-only`), `verify-docs-sync.py`, `verify-geometry.py --all`, `test-self-check.py`, `test-verify-geometry.py`, `test-verify-motion.py`, `test-verify-docs-sync.py`. Plugin manifests bumped together to 2.4.4.

End-to-end check: a 6-node Korean flowchart (Korean node names, Latin sublabels) built against the patched rules passes `self_check.py`, `verify-geometry.py`, and `lint-skin.py` clean, renders Hangul with no tofu, and shows no box overflow.

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples — no examples added or changed

## Related issues

None open for this. Filing one first if you'd rather review the approach before the diff — happy to split the measured-width rule out if you consider it a separate concern from the font stack.


etaseoul-jason:
Pushed a follow-up commit: the serif slot had the same hole as the sans one.

`Instrument Serif` has no Hangul, so a Korean `<h1>` fell through `Times New Roman` to the platform's generic serif - Apple SD Gothic Neo on macOS, Batang on Windows. Adding `Noto Serif KR` right after Instrument Serif in `--font-serif` pins it, and because fallback is per glyph, a mixed title keeps its Latin words in Instrument Serif.

Checked by glyph advance rather than by eye: in the patched stack `거` measures 61.83px at 64px, matching Noto Serif KR exactly (Noto Sans KR 58.89, Apple SD Gothic Neo 55.38), while `Rn` stays at 62.92px against Noto Serif KR's own 88.08px.

All gates re-run green. Happy to drop this commit if you would rather keep the serif slot Latin-only and treat Korean titles as the author's problem.

cathrynlavery (changes requested):
Please revise before merge. PR #83 has now landed the exact Korean proportional/mono fallback allowlist, so rebase onto current main and keep this PR focused on the always-loaded guidance, template font availability, and width budgeting. Greptile's current-head review found a real sizing gap: the formula counts Hangul, Latin, and spaces but ignores digits and punctuation, so common mixed labels such as `주문 v2.1` can still receive undersized boxes or masks. Extend the width budget to account for every rendered character class (with a conservative fallback), add adversarial mixed Hangul/Latin/digit/punctuation cases, run the complete current-main suite, and bump both manifests to the next unique synchronized version.

Issue #62, opened 2026-08-15
Proposal: new dataviz types — treemap, sankey, slopegraph + more


0xDarkMatter:
Love this repo @cathrynlavery ! I've been using some charts and dataviz in my own projects recently so thought I would contribute some extra chart types.

The current 27 types cover diagrams really well. On the chart side there's bar, line, and radar — these would add proportion (treemap), flow (sankey), rank change (slopegraph), paired values (dumbbell), part-of-whole counts (waffle), and more.

I've drafted 16 candidates in the house type-reference format, each with a working example in the default skin using public data (OWID, HadCRUT5, Olympic medals, US Census). They're all below — **pick whichever you'd like included**.

They all follow the house rules — one accent, ink-opacity ramps, mono for numbers, sourced footnotes — and each has an honest-data rule (treemap area is the only encoding, never clipped or log-scaled, etc).

If you're interested I'd PR whichever you choose, done per the contributing checklist: type reference, SKILL.md row, three static example variants, gallery entries, all gates green. My drafts have hover tooltips and a theme toggle for easy review but I'd ship fully static per ADR 0001.

Full drafts (type specs + honest-data rules): [dataviz-types.md](https://github.com/0xDarkMatter/diagram-design/blob/dataviz-preview/docs/proposals/dataviz-types.md) — each image below links to the live HTML.

| | |
|---|---|
| **Treemap** — hierarchical proportion<br>[![Treemap](https://raw.githubusercontent.com/0xDarkMatter/diagram-design/dataviz-preview/docs/proposals/dataviz/shots/world-population-treemap.png)](https://raw.githack.com/0xDarkMatter/diagram-design/dataviz-preview/docs/proposals/dataviz/world-population-treemap.html) | **Sankey** — flow between two axes<br>[![Sankey](https://raw.githubusercontent.com/0xDarkMatter/diagram-design/dataviz-preview/docs/proposals/dataviz/shots/migration-sankey.png)](https://raw.githack.com/0xDarkMatter/diagram-design/dataviz-preview/docs/proposals/dataviz/migration-sankey.html) |
[...]

PR #29, opened 2026-08-12
docs(contributing): add CONTRIBUTING, code of conduct and issue/PR templates

philyuchkoff:
docs(contributing): add CONTRIBUTING, code of conduct and issue/PR templates

Add `CONTRIBUTING.md` covering all validation gates and contribution workflows, `CODE_OF_CONDUCT.md` (Contributor Covenant 2.1), GitHub issue templates (bug/feature), and a pull request template; add a Contributing section to the `README`.

PR #87, opened 2026-08-18
feat(types): add treemap for part-of-whole by area

0xDarkMatter:
Adds Treemap as the 28th visual type. Closes the first item in #62.

**Why a top-level type and not a variant.** ADR 0002 holds the count unless a genuinely new *layout* grammar appears. Recursive area subdivision is one — bar encodes with length, nested with containment and no quantity, pyramid with rank. Most of the other candidates in #62 genuinely are variants, which is why this PR is only the treemap.

## Three decisions worth your attention

- **The focal cell uses tint-plus-accent-stroke, not a solid accent fill.** Solid `#eb6c36` is a mid-tone that clears neither white (3.1:1) nor ink (3.6:1) at 4.5:1, so the 9px value labels would fail either way. Related: the fill ceiling is computed against `muted`, which is what the value line actually uses — `ink` would tolerate 0.20, `muted` needs 0.16 light / 0.14 dark. Worst text contrast now measures 4.57 and 4.64.

- **The smallest cell carries an info mark, not a label.** Oceania is 0.6% of the world; a rotated label would be the only sideways text in the figure, so it gets an ink disc with an `i` and is named in the legend. The cell is still drawn at its honest area — a treemap that omits a part stops being a picture of the whole.

- **`scripts/verify-treemap.py` is new, and I'd understand if you want it separate.** It gates the two things a treemap breaks silently: a cell's drawn share against the value printed inside it, and every label against the cell holding it. It exists because an earlier cut of this diagram drew Oceania 24% under its true share while sitting comfortably inside a 1-percentage-point tolerance — absolute error is blind to exactly the cell most likely to be wrong, so it measures relative error. Pure Python, no browser, wired into CI with adversarial tests both ways per ADR 0005. Happy to pull it into its own PR if you'd rather keep this one to the type alone.

## Two things I touched beyond the type

Both because this PR would otherwise break them:

- ADR 0002 names the two counting verifiers as its own enforcement, so moving them 27 → 28 without amending it would quietly make the contributor the authority. It gains an amendment recording the treemap decision.
- The four plugin-manifest descriptions repeat the SKILL.md description verbatim and had drifted out of lockstep. Fixed, and `verify-docs-sync.py` now asserts it so type #29 can't break it silently.

## Shipping set

The full §10 set: three variants (dark and editorial generated from the light one so their geometry can't drift), type reference, gallery tab, §3 row, description hook, §7 complexity budget, README entry and screenshot.

**Tested:** every gate in CONTRIBUTING plus the two new ones — 18 green. Packaged `self_check.py` OK on all three variants; rendered light, dark and editorial to confirm.

Minor bump to 2.5.0, following #61's precedent for a feature.


0xDarkMatter:
Good catches, both fixed in 34a86a6. The reference genuinely did contradict itself — the honest-data rule still described the rotated label the layout section had already replaced with an info mark. And the fit check was only testing right/bottom edges; fixing that surfaced a related bug where an end-anchored label's box corner could resolve to the wrong host cell. All four edges are checked now, with adversarial cases for the two new directions.

cathrynlavery (approved):
Current-head Greptile review is 5/5 after the sliver-marker containment fix, the rendered treemap is visually coherent, all six cell areas match their stated shares within the documented tolerance, both manifests are synchronized at 2.5.0, and the complete local and current-head CI suites are green.

samariyan60-a11y:
https://t.me/AI_in_Research/5286

Issue #78, opened 2026-08-17
[Feature]: ten figures the current 27 types cannot draw (four types, six variants)


JacopoVendramin:
### What are you trying to show?

I ran a coverage audit of the 27 types against the Mermaid 2026 taxonomy and a list of figures
engineers had actually asked me for, then filtered the results through the ADR 0002 bar: a new
type is only justified when a **layout grammar** is missing, never when a new *use case* appears.

Most of what the audit surfaced failed that bar and is already served: system context, component
and API-interaction diagrams are Architecture at a different zoom; request lifecycles are Sequence;
DFDs are Data Flow; decision trees are Flowchart; conceptual data models are ER.

Ten figures survived, and they split into two groups. **Four** need a grammar no existing type can
represent — not renders differently, structurally cannot represent. **Six** are real gaps that
extend an existing grammar rather than replacing it, so I'm proposing those as variants.

All ten are implemented and passing the full gate suite, so this is a review question rather than
a feasibility one. I've read #55, #62 and #63 and tried to apply the standard those set.

---

### Proposed as new types

Each row names the capability the nearest existing type lacks.

| Type | What it shows | Nearest existing type, and the specific thing it cannot do |
|---|---|---|
| **Sankey** | A quantity splitting and merging across stages | Band width *is* the datum. Pyramid shows drop-off but not splits and merges; Process shows ordered steps but carries no magnitude. No existing type encodes quantity as area. |
| **Fishbone / Ishikawa** | Causes of one observed effect, grouped by category | Tree is parent→children and **diverges**. A fishbone **converges**: angled category bones meeting one effect at a fixed angle, with sub-cause ticks along each bone. Tree has neither convergence nor bone geometry. |
| **Dependency graph** | What depends on what, across packages or services | Tree structurally forbids the two things that make this diagram worth drawing: a node with **more than one parent**, and a **cycle**. Fan-in convergence and a back-edge are exactly what a reader looks for here. |
| **UML class** | Classes with operations, inheritance, composition | ER has no operations compartment, and no vocabulary where the **arrowhead carries the semantics** (hollow triangle vs filled diamond vs hollow diamond). Crow's-foot cardinality answers a different question. |

**On Sankey specifically.** #62 proposed it and you named it the next candidate after Treemap, so
@0xDarkMatter has first claim. Mine is built and passing, and I'm putting it forward rather than
sitting on it, but if that work is already moving I'll drop mine without argument — say the word
[...]

Issue #63, opened 2026-08-15
[Feature]: cloud architectures need a containment grammar (an Azure case study)


msouga:
### What are you trying to show?

I want to show an Azure architecture where **the network hierarchy itself carries
the meaning**, and the reader needs to know what sits inside which subnet.

`type-architecture.md` is provider-agnostic, which is the right default. But it
caps at three zones, flat, with no nesting, and its examples label them `EDGE`,
`CORE SERVICES`, `DATA`. A real Azure topology has four levels of containment
that are not interchangeable:

```
subscription
  └── resource group
        └── virtual network
              └── subnet
                    └── the nodes
```

Drawing this today means flattening two levels into one label. I ended up with a
zone eyebrow reading `VNET · SNET-APP`, which passes every gate and still lies
about the hierarchy.

The distinction is load-bearing, not cosmetic. A resource group holds no
traffic — it groups resources for billing, permissions and lifecycle. A virtual
network is a boundary traffic does not cross without an explicit rule. A reader
who cannot tell those apart cannot tell whether a design is secure.

There is a second thing the current type cannot express, and it is the error I
see most often in Azure diagrams in the wild: **a PaaS service is never inside a
subnet.** Azure SQL, Key Vault and Storage belong to the resource group; what
lives in the subnet is their *private endpoint*. Diagrams that put the database
inside the subnet tell the reader it sits behind that subnet's NSG, which is
false.

### Proposed solution

I have a working implementation and I would rather ask before polishing it for a
PR, because the shape of the answer changes the work.

**The open question is the same one as #55: is this a new type, or a variant of
an existing one?** I can see it either way:

- **A new `type-architecture-azure.md`**, extending `type-architecture.md`. All
  of that type's rules still apply; the new file adds only the containment
  grammar. That is what I built.
- **A containment primitive** that `type-architecture.md` gains, usable by any
  cloud. AWS has the same shape (account → VPC → subnet) and, as far as I can
  tell, is equally unserved today — the repo's own showcase example draws an AWS
  architecture with the generic grammar.

What the implementation does, in case it is useful either way:

- **Reuses existing vocabulary rather than inventing any.** Containment geometry
  comes from `type-nested.md`: consistent insets, eyebrow per level, fills
  stepping up inward, strokes hardening. Administrative levels render as
  horizontal bands, which `type-swimlane.md` and `type-layers.md` already
  establish.
- **Adds two layout rules** that fell out of drawing it, both about density:
[...]

PR #98, opened 2026-08-18
feat(types): add database schema as a physical ER variant

JacopoVendramin:
Part of the split you asked for in #79. One grammar, one product decision, cut fresh from current `main`.

## Why this one clears the bar

A database schema draws the **physical** model: foreign keys anchored **column row to column row**, SQL types, constraint chips, indexes, schema grouping and delete behaviour. ER draws the conceptual and logical model, where none of those exist.

| Nearest existing type | Why it does not cover this |
|---|---|
| ER / data model (as-is) | Relationships connect entity boxes and carry cardinality. A physical schema needs the edge to land on the exact column pair, because which column carries the key is the fact being stated. It also has no vocabulary for SQL types, indexes, or ON DELETE. |
| Nested | Schema grouping only. No columns, no keys. |
| UML class | The operations compartment is meaningless for a table, and the relationship vocabulary is ownership, not referential integrity. |

Column-level anchoring is the grammar. Once the edge lands on a column row, the diagram can state the delete behaviour on that specific constraint, which is the thing that actually bites in production.

## What ships

- `references/type-db-schema.md` — the house-format reference: table geometry, constraint chips, the index block, column-level FK anchoring, and delete-behaviour labelling.
- Three static variants: minimal light, minimal dark, full editorial.
- Gallery entry, §3 selection-table row, description hooks in all five slots, README preview cell with a rendered PNG.
- ADR 0002 amendment. #87 added the rule that a PR moving the counters must amend that record in the same PR, so this one does.

Counters move 28 → 29 and the manifests bump to 2.6.0, synchronized.

**One change to an existing file.** `type-er.md` said "Best for: database schemas". With a column-level physical variant in the taxonomy that leaves two routes competing for the same request, so ER is now scoped to the conceptual and logical model and points here for the physical one. Without that edit the router is ambiguous, so it belongs in this PR rather than a follow-up.

## Connector rules

Obeys §6 in full: orthogonal elbows, bridge or hop at every crossing.

## Validation

All green on this branch: `verify-docs-sync.py`, `verify-semantic-motion.py` (both modes), `verify-motion.py --shipped`, `verify-geometry.py --all`, `lint-skin.py --all --baseline`, `verify-plugin-package.py origin/main`, `verify-drawio-import.py`, `verify-mermaid-import.py`, `verify-sequence-oauth.py`, `verify-treemap.py --all`, plus the `test-*` set.

No new entries in `lint-skin-baseline.txt`. The three variants were rendered and reviewed, not just linted.

## On the split

Since I cannot open stacked PRs from a fork, each item in the split independently claims count 29 and version 2.6.0 at open time. Whichever you take first keeps those numbers and I renumber the rest against `main` — same collision treemap and #79 just had, and no ordering is implied. Sankey and the deployment variant are excluded entirely, per your ruling: they stay with #62 and #63.

Refs #78, supersedes part of #79.


cathrynlavery:
Thank you for this awesome diagram. We're consolidating this and the rest of the set into #118 so all ten types ship in one plugin version and one coordinated docs/gallery update, without version and count conflicts between the focused PRs. Your authorship and follow-up fixes are preserved in the integration branch.

We'll close this PR as superseded after the integration passes its current-head checks and merges.

cathrynlavery:
Closing as superseded by #118, which shipped Database schema as a canonical type and preserved your contributor-authored work. The physical-table grammar—including SQL types, constraints, indexes, and column-level foreign keys—is in the merged release. Thank you, @JacopoVendramin.

PR #102, opened 2026-08-18
fix(treemap): verify every cell's area, including cells too small to label

### LINKED ISSUE #89: Treemap verifier skips unlabelled sliver area and 2.5 metadata drift
## Problem

PR #87 added Treemap and its verifier, but a post-merge sweep found three gaps on current main:

1. `verify-treemap.py` normalizes only cells with in-cell numeric labels. Oceania carries only an info mark and its value is in the legend, so changing its cell from 16×124 to 24×124 (a 50% distortion) still reports the treemap clean.
2. `skills/diagram-design/SKILL.md` still declares `metadata.version: "2.4"` while both plugin manifests are 2.5.0.
3. The displayed parts total 8.10B while the source says 8.09B, but the examples omit the reference-required `PARTS ROUNDED, MAY NOT SUM` disclosure.

## Acceptance criteria

- Add machine-readable value/share metadata to every treemap cell, including unlabelled slivers.
- Verify every cell against that metadata; fail closed when an expected treemap has too few parseable cells.
- Add adversarial oversized and undersized unlabelled-sliver tests that fail on current main.
- Synchronize SKILL metadata at 2.5 and extend the package gate so future minor releases cannot drift.
- Add the rounded-parts disclosure to all three shipped variants.
- Run the complete repository suite and bump both plugin manifests to the next unique version.

0xDarkMatter:
Closes #89. All three were real — good catches, especially the first one.

**Sliver gap.** The area check took its basis from in-cell labels, so a cell without one dropped out of the checked set entirely. Repro'd on main: widening Oceania 16px → 24px, a 50% distortion, still reports `OK`.

Every cell now declares `data-share` (its percentage of the whole) and the verifier checks all of them against it. Missing metadata fails closed instead of skipping. It also cross-checks the label against the metadata — they're two statements of one fact and can drift apart.

Tests: +4 cases — oversized and undersized unlabelled sliver (both pass on main, fail here), missing metadata, label/metadata divergence. 12 total.

**Version drift.** `SKILL.md` said 2.4 against 2.5.0 manifests: a script bumps one, the other is hand-maintained, nothing read both. `verify-plugin-package.py` now asserts SKILL.md tracks the manifest MAJOR.MINOR. The package test's fixture gains the field too, since a SKILL.md without one isn't a valid package.

**Disclosure.** `PARTS ROUNDED, MAY NOT SUM` restored to all three variants — lost when the examples were rebuilt around the legend block. Fixed a contradiction next to it while I was there: the anti-pattern list forbade outright what the honest-data rule permits once disclosed.

One cosmetic change rode along — the info mark now sits on the label row of its neighbours rather than its own cell's centre, so the bottom row reads as one row.

Manifests at 2.5.2, 18 gates green locally.


cathrynlavery:
Thanks for contributing — your focused implementation was the foundation for the current-main integration in #124. I cherry-picked your commit with authorship preserved, reconciled it with the newer CJK treemap verifier work, and completed the remaining fail-closed cases identified in review: invalid/non-finite/out-of-range shares, conflicting or extra duplicate cells, insufficiently parseable treemaps, and strict frontmatter-only `metadata.version` parsing. The full repository suite now passes on the integrated branch. Because #102 conflicts with the releases that landed after it, #124 is the coordinated merge path; no further changes are needed from you. I appreciate the careful reproduction and adversarial tests.

cathrynlavery:
Thank you again for the original treemap verification work. Your authored commit was preserved in #124, and the post-merge adversarial hardening completed in #127. Both are now on main, issue #89 is closed, and the final package is at 2.5.18. I’m closing this older PR as superseded; no further action is needed from you.

PR #103, opened 2026-08-18
Add sankey for flow that splits and merges across stages

0xDarkMatter:
Love a good sankey. Here's the one from #62.

Treemap landed while I was building this, so it's rebased on #87 and the count goes 28 → 29. Full §10 set, three variants, gallery tab at 30 right after Treemap's.

Heads up there's a new checker in here too.

### Why this is a type and not a pattern

ADR 0002 says the count moves only for a genuinely new *layout* grammar, so here's the argument explicitly. Sankey's grammar is **flow-conserving ribbons between staged nodes**: width is a volume, that volume survives from one stage to the next, and a node is where volumes merge and split. The nearest candidates each fail on the same axis — pyramid/funnel narrows monotonically and cannot branch or merge, bar encodes with length against a shared baseline rather than width along a path, tree draws parent-to-child structure with no quantity, and treemap encodes a single moment by area with no movement between states. None can express conservation, which is the entire claim a sankey makes.

ADR 0002 is amended here rather than after, layered under Treemap's amendment. That record names `verify-semantic-motion.py` and `verify-docs-sync.py` as its own enforcement, so a PR that edits those counters without amending it has quietly appointed itself the authority.

### N-stage, not just source → target

Numbered stage headers, thin node bars rather than boxes, optional node sublabels, a legend naming the ribbon classes, and ribbons carrying across multiple hops with an optional inline annotation. Capped at four stages and at one accent plus two ink steps.

Three deliberate departures worth calling out:

- **One accent plus at most two ink steps.** A third ink step inside the contrast ceiling is indistinguishable in a 16×10 legend swatch, which makes it decoration.
- **Tone never encodes volume.** Width already carries it; a second encoding of the same quantity is ink a reader wastes time decoding.
- **Sublabels are optional and most nodes don't get one.** The example runs five on eleven. A sublabel that restates its node name is the name twice.

### The example

A four-step signup funnel — 12,000 starts, four traffic sources into two devices into three form outcomes into activated/lost. Illustrative product data, in the register the other chart examples use, and the source line says it's illustrative rather than dressing invented numbers in a provenance they don't have.

A funnel is a deliberate choice of subject: it's where drop-off gets hidden, so it's the sharpest test of the rule that every session leaving the story lands in a named node and every stage carries the same 12,000.

### The gate

1. **Conservation** at every node. A ribbon quietly narrowed between two stages is invisible — the eye has no reference width, so 7,200 arriving and 7,000 departing looks exactly like 7,200 and 7,200.
2. **Stage totals** against each other, which catches a whole node being forgotten.
3. **Ribbon-width constancy** — the ends are hundreds of pixels apart, so a taper reads as perspective. It isn't.
4. **Value fidelity** — drawn height against printed value, one scale, relative error.
5. **Control-point placement.** B′(1) = 3(P₃−P₂), so the arrival tangent is set by the last control alone. Both on the corridor midline and the ribbon plugs into its bar flat; one on the target's x and it coincides with the endpoint, the first derivative vanishes, and the visible approach falls to B″(1) — 133° off horizontal here. Every endpoint stays correct, so nothing else would catch it.

With `--all`: all three variants must be present and geometrically identical, and the reference's element pattern must be a **literal excerpt** of the shipped example — a code sample is the one piece of geometry nothing renders, so it rots silently the first time the layout moves, and it's exactly what the next contributor copies.

Guards underneath, because the worst outcome for a checker isn't a missed defect but a clean exit having read nothing: an unparseable `<path>`, a ribbon end attached to no bar, node bars outside the size window, a figure with bars but no ribbons, and a glob matching no files are all findings rather than silent skips. Widest of them runs first — a file carrying a `transform`, `clip-path`, `mask` or hidden element is **refused, not checked**, because any one breaks all five invariants while the numbers the checker reads stay correct, and catching that properly would mean reimplementing a browser. It costs nothing: no shipped sankey uses any of them.

`scripts/test-verify-sankey.py` covers **sixteen cases in both polarities** per ADR 0005, five of which attack the checker rather than the figure.

### Contrast

Computed against the token actually shipped. The value line is `muted`, not `ink`: over the composited fill it holds 4.5:1 up to 0.14 on light paper and 0.12 on dark. Two 0.14 ribbons stacked measure 3.74:1, so **no label sits where ribbons overlap** — verified analytically across all 33 labels in both minimal variants, not by eye.

### Validation

Every gate in CONTRIBUTING's table including Treemap's two, plus `self_check.py` on all three variants, plus the two new ones — 20/20. SKILL.md lands at 37,921 bytes against the 40,000 cap. Version bumped minor to 2.6.0 per #61.

### Not included

No Mermaid or draw.io import path. `sankey-beta` sits in the extractor's `UNSUPPORTED_KINDS` alongside `gantt`, `quadrantchart` and `timeline`, which also have visual types — only four grammars have importers today, and adding a fifth is its own change.

---

Happy to change any of it — the four-stage cap, the accent going to the outcome rather than the drop-off, the example subject.

Shout if you'd rather the checker was its own PR and I'll split it.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, DP security matrix, sankey, fishbone, Wardley map, kanban, user journey, deployment, dependency graph, UML class, story map, or database schema diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.5.20",
+  "version": "2.5.21",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, DP security matrix, sankey, fishbone, Wardley map, kanban, user journey, deployment, dependency graph, UML class, story map, or database schema diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.5.20",
+  "version": "2.5.21",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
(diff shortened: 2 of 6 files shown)


cathrynlavery:
Thanks for the Sankey work here. Sankey itself has now shipped as a canonical type in #118, so the type/reference/examples/gallery/version portion of this branch is superseded. The validator and adversarial tests remain useful. Please rebase onto current `main` and revise this PR to a validator/tests-only change that checks Sankey quantity conservation and ribbon geometry against the shipped examples, with no duplicate type files, routing/count edits, gallery registration, or plugin-version claim. Once that focused diff passes current CI, we can review it on its own merits.

cathrynlavery:

cathrynlavery:

cathrynlavery:
@0xDarkMatter, thank you for adding Sankey support. Flow conservation and ribbon geometry are exactly the details that make this visual trustworthy, and your contribution gave us a solid foundation for diagrams that split and merge across stages. I am delighted to have it in the release.

0xDarkMatter:
Thanks @cathrynlavery ! Happy to contribute and glad you like the Sankey

PR #106, opened 2026-08-18
feat(line): add slopegraph variant for change between two states

0xDarkMatter:
Slopegraph as a Line variant, not a new type — third from #62. Count stays at 28.

- `type-line.md` gets a `## Variants` section
- Three examples, gallery tab 30
- New gate `verify-slopegraph.py`: both axes must share one scale, and every
  endpoint sits where its printed value belongs. 38 tests
- One SKILL.md cell so "slopegraph" routes to the reference — drop it if you'd
  rather

```
export PYTHONIOENCODING=utf-8
python scripts/verify-slopegraph.py --all
python scripts/test-verify-slopegraph.py
```

Example is p95 latency per service before/after a cache, one 100–550ms scale on
both axes, illustrative numbers.

Pre-existing, left alone: `example-line-dark.html` isn't dark (1.67:1 text);
`example-treemap-dark.html`'s "darker is larger" is backwards on dark paper;
`accent` is 2.86:1 on light paper against 1.4.11's 3:1.

Gates pass. Bump next?

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, or DP security matrix diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.5.5",
+  "version": "2.5.6",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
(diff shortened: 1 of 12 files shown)


cathrynlavery (changes requested):
I reproduced four current-head correctness/safety defects at `d8139c7f27ce7bfa475dcebcde58a185d1e85416`; please address them before merge.

1. `scripts/test-verify-slopegraph.py` writes fixtures to fixed paths under the repository root and unconditionally unlinks them. An existing `.slopegraph-fixture.html` (or other selected fixture name) can be overwritten and deleted, and concurrent runs race. Use an isolated temporary directory/file and add a regression test proving a pre-existing same-named file is preserved and parallel runs cannot collide.
2. The verifier validates raw `x`/`y` attributes but ignores SVG/CSS/ancestor transforms. I added `transform="translate(0 80)"` to a series line and the verifier still reported OK, although the rendered endpoint no longer matches its printed value. Either reject transforms on verified geometry/ancestors or resolve them into rendered coordinates; add adversarial regression coverage.
3. Printed values are parsed with a first-fragment regex. Changing a visible `512` label to `512,000` still reports OK against metadata value `512`. Parse and validate the complete visible numeric token (including the supported thousands/decimal policy) and add malformed/suffixed/formatted regression cases.
4. The gate does not bind visible state captions or series names to the semantic metadata. Swapping `BEFORE` and `AFTER` still reports OK, so a semantically reversed chart can pass. Validate those visible labels and add swapped-caption/name regressions.

Completion means these regressions fail against this head and pass with the fix, the complete local suite and render checks pass, all conversations are resolved, and exact-head GitHub CI plus a fresh Greptile 5/5 are green. If `main` advances first, rebase and use the next unique synchronized package version before obtaining those final-head results.

0xDarkMatter:
All four reproduced at `d8139c7` and are reported at `0bc07b3`.

1. Fixtures now live in a per-process temp dir. Two cases hold it: a sentinel
   file at the old repo-root path survives a full run, and two harnesses never
   share a directory.
2. Transforms rejected on series lines, bound labels, ancestor `<g>`/`<svg>`,
   and in CSS. Not resolved — a partial SVG transform stack looks like coverage
   without being it. The rotated value-axis caption still passes, and
   `text-transform` is not `transform`.
3. The complete visible token is parsed, thousands separators included. A unit
   suffix (`208ms`) stays legal; a second number in one label is reported
   rather than half-read.
4. Captions carry `data-axis` + `data-state`, names carry `data-role="name"`,
   and each visible string is cross-checked against its binding. Also reported:
   any label drawn nearer another series' endpoint than its own.

On 4 — my first version used the *nearest* endpoint and wrongly reported two
series that legitimately share a value, and every from-endpoint of an indexed
slopegraph. Ties pass now.

Tests 38 → 58. type-line.md documents the binding contract as a table of what
each attribute buys and what omitting it costs.

Rebased onto `2991772` (dumbbell), version 2.5.6. One thing unrelated to this
PR: #107's run-all block in CONTRIBUTING.md landed with literal `\n` instead of
line breaks, so the documented command doesn't paste. It was inside the lines I
had to resolve, so it's repaired here — `bash -n` parses the block now.

PR #100, opened 2026-08-18
docs(high-level): register datalake examples as unclustered High-Level variants (fixes #55)

philyuchkoff:
Closes #55 per maintainer decision.

datalake examples (example-datalake.html, -datalake-dark.html, -datalake-full.html) are registered as unclustered High-Level variants — the dominant grammar is an ordered five-phase data stack without a container-orchestrator boundary. The MinIO lake is the focal node; vertical concerns are flattened into the horizontal chevron banner. No type count change.

- Added three example entries to `references/type-high-level.md` §9

PR #107, opened 2026-08-18
feat(bar): add dumbbell as a Bar variant

0xDarkMatter:
Dumbbell as a Bar variant, per your call on #62.

Reference-only: the existing variants in `type-bar.md` are prose-only with no assets, and #100 landed datalake the same way. 3 files, +61/-3. Count stays 28 — SKILL.md, the counters and the gallery all untouched. Manifests 2.5.2 → 2.5.3.

Layout on the 1000×500 composition, an element pattern, and five honesty rules: no truncated axis, one scale with both endpoints labelled, connector is a gap not a trajectory, stated row order, no imputed or silently dropped endpoints.

No gate — with no shipped example a checker would read zero files and pass. Worth one when a dumbbell example lands.

Built light and dark charts from the section text alone to check it, a decreasing row and a zero-floor row included; both passed `lint-skin` and `self_check`. All 18 gates green.

Shout if you want it shaped differently.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, or DP security matrix diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.5.4",
+  "version": "2.5.5",
   "author": {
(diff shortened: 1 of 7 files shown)


cathrynlavery (changes requested):
Two merge-blocking gaps remain at `e970245`:

1. The sign-based domain rule does not define an all-zero dataset (and is ambiguous for data that only touches zero). The documented formula then reaches `floor == ceil == 0` and divides by zero, so valid zero-gap comparisons cannot be positioned. Define zero-inclusive cases explicitly and require a finite, non-zero fallback domain for all-zero data.
2. The reference calls the solid endpoint and connector load-bearing while documenting 2.86:1 and 1.60:1 contrast against paper. Shape redundancy does not make the solid mark's boundary or the connector compliant. Specify a compliant stroke/direct-label treatment so understanding the two series and pairing does not depend on sub-3:1 marks.

Please add regression coverage that exercises all-zero and zero-touching domains through the actual scaling/checking path and proves finite coordinates, plus an adversarial contrast case for the chosen endpoint/connector treatment. A prose-only assertion is not enough for these formula-level rules.

Completion condition: resolve the two defects and tests, rebase onto current `main` if another version lands first, bump both manifests to the next unique synchronized version, pass the complete local suite and exact-head GitHub CI, resolve every thread, and obtain a fresh Greptile 5/5 on that same final head.

PR #112, opened 2026-08-19
fix(commands): stop hardcoding the visual-type count, and gate it

mysoniw:
## What does this PR do?

Both import commands said `--type` forces one of the **27** while the selection table was at 28 (treemap, #87); today they say 38, freshly edited by #118. The number is wrong in a particular way: it has to be edited by every PR that adds a type, and is the one file such a PR has no reason to open. #109 fixed the drawio *reference* the same way and left this line behind.

Bumping the number goes stale again as soon as the next type merges — #118's own release bump to 38 shows the sentence has to be found and edited every time. So the commands point at `SKILL.md` §3 instead — where the answer already lives, and where it is already gated, since `verify-docs-sync.py` checks the selection table against the frontmatter description on every run.

`verify-docs-sync.py` gains `check_type_counts` so this class of drift cannot come back.

```
commands/import-mermaid.md:25 hardcodes the visual-type count; point at SKILL.md §3
instead so adding a type cannot leave it stale
```

## Validation gates

- [x] `python3 scripts/verify-docs-sync.py && python3 scripts/test-verify-docs-sync.py`
- [x] Full local suite green (28 gates) after rebasing onto the 38-type `main`; Claude, Codex, and Factory manifests bumped to 2.5.16

## Checklist

- [x] Generated and source files are consistent (checker ↔ test ↔ command ↔ docs)
- [x] Accessible SVG contract satisfied for new/changed examples — none changed

Adversarial coverage in `test-verify-docs-sync.py` runs all three polarities: a command that points at SKILL.md passes, one that spells a number out fails, and a missing command surface is reported rather than skipped.

## Related issues

None. Noticed while reading the import commands.


mysoniw:
Rebased onto `main`. Both review comments are fixed:

- The count gate matches the count itself rather than one phrasing — `one of the 27`, `28 visual types`, `28 types of visual diagrams` — with whitespace-tolerant gaps, since both commands wrap that sentence and a count can land just after the wrap.
- The missing-surface test no longer carries the previous case's fixture, and asserts the exact error instead of `any(...)`.

Also fixed one I introduced while doing that: the first broadening rejected `accepts 2 file types`. The pattern now requires the taxonomy noun in both word orders.

Word-form numerals are out of scope — README and the docstring say "numeral" so the gate doesn't claim more than it checks.

cathrynlavery:
Thanks for tackling the hardcoded-count drift. PR #118 moved `main` to 38 types, and this branch is now based on the pre-release count and version. Please rebase onto current `main` and keep this contributor-owned PR focused on the count-free documentation gate: derive and validate the wording structurally instead of replacing `37` with another literal `38`, and drop changes now made redundant by #118. After rebasing, bump from current main to the next unique synchronized patch version and rerun the full package and docs gates. I'd like to preserve this PR's ownership rather than recreate the fix elsewhere.

mysoniw:
Rebased onto current main and bumped to 2.5.16 (the new Factory manifest included).

The gate is still count-free. The rebase pulled in #118's literal `38` on those two command lines, and I resolved it back to the SKILL.md §3 pointer, so nothing in this PR knows what the count actually is. I also went through the diff and dropped what #118 already covers — what's left is the two command lines, the gate with its tests, and a sentence each in README and CONTRIBUTING.

While I was in there I noticed the checker's module docstring enumerates the drift classes and didn't mention the one this PR adds, so that's counted now too.

Everything passes locally, 28 gates, including verify-plugin-package against current main.

cathrynlavery:

cathrynlavery:

cathrynlavery:
@mysoniw, thank you for removing the hardcoded visual-type count and adding a gate to keep it from drifting again. That makes future additions safer and keeps the routing guidance honest as the library grows. This is exactly the kind of maintenance improvement that compounds over time.

PR #113, opened 2026-08-19
feat(mermaid): support gantt, quadrantChart, timeline, and mindmap grammars

mysoniw:
## What does this PR do?

Teaches `mermaid_extract.py` four grammars it currently rejects: `gantt`, `quadrantChart`, `timeline`, and `mindmap`.

The gap was a parser, not a taxonomy question. Each of the four already has a shipped visual type to redraw into, so **no new type is added and the 38-row selection table is untouched**:

| Grammar | Target type |
|---|---|
| `gantt` | Gantt |
| `quadrantChart` | Quadrant |
| `timeline` | Timeline |
| `mindmap` | Tree |

Each carries structure that is neither a node nor an edge — a date format, axis and quadrant-cell names, a declared title — so `Diagram` gains a `meta` mapping that the digest and the JSON both report. Two smaller consequences: the `budget:` line now names the cap it applied, because SKILL.md §7 gives Gantt tasks and quadrant items 12 rather than the default 9; and grammars that are edgeless by design omit the `unconnected:` line instead of listing every node there, which would be noise rather than the first rung of the degrade ladder.

Three placement decisions worth stating, since they are not obvious from the source:

- **A gantt positional field is a date only when it is digits and separators.** Otherwise `:crit, dual, after inv, 15d` reads its own task id as a start date.
- **Timeline events stay as fields of their period.** Promoting them to child nodes turns every drawn period into a container, which the drawable-node budget then stops counting.
- **Mindmap branches become parent → child edges, not container membership.** Every topic in a tree redraw is drawn, so none of them should be counted as grouping.

The trust boundary is unchanged: nothing is rendered, fetched, or executed, and every label stays inert text. The `mindmap` parser deliberately skips the shared style-directive filter — `style` is not a mindmap keyword, and a topic that begins with it must not vanish. There is a fixture for exactly that, alongside markup, a URL, and an injection string.

## Validation gates

- [x] Full local suite green (28 gates) after rebasing onto the 38-type `main`; earlier rounds also ran on Python 3.10/3.11 with syntax verified on 3.8
- [x] Claude, Codex, and Factory manifests bumped to 2.5.17

## Checklist

- [x] Generated and source files are consistent (extractor ↔ verifier ↔ reference ↔ command ↔ prompt)
- [x] Accessible SVG contract satisfied for new/changed examples — none changed

Per CONTRIBUTING, the new grammars come with fixtures and verifier coverage in the same change: five fixtures under `scripts/fixtures/`, five new checks in `verify-mermaid-import.py`, and the unsupported-kinds assertion now derives from the extractor's own `SUPPORTED_KINDS` rather than a second hardcoded copy that had already drifted once.

## Related issues

None open for these grammars.


mysoniw:
Rebased onto `main`. Half the review comment is fixed, half I think is not a defect, and checking it turned up three more.

**Fixed — continuation lines.** A leading-`:` line now attaches to the period above it instead of drawing an unlabelled period. With no period above it, exit 2 rather than inventing one.

**Declined — colons inside label text.** Mermaid's timeline treats every colon as an event separator (`2004 : Facebook : Google` is two events) and has no escape syntax; the limitation is tracked as mermaid-js/mermaid#6557 and #5699. Splitting on every colon is what matches Mermaid. Correct me if I have the grammar wrong.

**Three more, all in code this PR adds:**

- Quadrant accepted any coordinate. Now bounded to the unit square the reference promises.
- Gantt classified metadata by inspecting values, which cannot work — `dateFormat` decides what a date looks like (`Jan` under `MMM`, `PM` under `A`), so dates in those formats were silently dropped. Now read by item count, the way Mermaid's own table is defined. That also fixes a lone date being labelled `start`, which drew single-date tasks a whole task-length early.
- Arity the table has no row for — a fourth item, a tags-only task, an empty slot — now exits 2 instead of being rounded down to something plausible.

Each failure path is covered in `verify-mermaid-import.py`, and `references/import-mermaid.md` documents the new exit-2 cases.

mysoniw:
Rebased onto current main and bumped to 2.6.2. The version this branch carried predated 2.6.0, so it sat below main rather than above it.

Checking the rebase turned up three ways this branch's extractor returned fewer nodes than the source declared, with no error and no discard counter.

A generated id is numbered by node count, so a source that declares `task-2` or `topic-2` collides with one, and `add_node` merges on a repeated id. A mindmap whose root is written `topic-2((Root))` came back as one node with a self-edge and `cycle: True` instead of a three-node tree. Generated ids now skip past anything already declared.

`_discard_nonsemantic` matched a bare `style `/`click ` prefix, so a quadrant point named "Style debt" and a gantt task named "Click rate" were dropped as directives. Quadrant now reads a line as a point before considering it a directive, and gantt matches Mermaid's actual `click <task> call|href` form — gantt has no style statement to match in the first place. A real `click a1 href "..."` is still discarded and counted.

Both are covered in `verify-mermaid-import.py`, and the new cases fail against the previous extractor.

I also narrowed two claims in the reference that the code doesn't enforce: there is no mindmap depth budget, and gantt omits `unconnected:` whether or not `after` dependencies exist.

mysoniw:
The timeline colon note is the one I declined earlier and still think is correct: Mermaid's timeline treats every colon as an event separator and offers no escape, so `2004 : Facebook : Google` is two events there too. Tracked upstream as mermaid-js/mermaid#6557 and #5699. Splitting on every colon is what matches the renderer; diverging would make the redraw disagree with the source. Happy to change it if I have the grammar wrong.

mysoniw:
Another review pass turned up seven places where the extractor didn't follow the grammars it claims to support. I checked each against Mermaid's syntax docs and reproduced each one before changing anything.

Four lost or corrupted content silently. My earlier id-collision fix only skipped ids that already existed, so a declared `topic-2` reached *after* an anonymous sibling still merged the two — the ids a block declares are now collected before parsing. A mindmap Markdown string spanning several lines became three topics with the delimiters stranded in their labels. Repeated `excludes` lines overwrote each other where `docs/syntax/gantt.md` says the tokens are concatenated. A quadrant point's `:::class` stayed in the label instead of being counted as discarded styling.

`weekend friday` is a valid gantt directive and was rejected as a malformed task.

`timeline TD` was ignored and always reported `LR`, and a timeline row with no event became a period node — inventing a marker the source never described. That one exits 2 now, which is separate from the colon-splitting question above: this is about a row with no colon at all.

Each has a case in `verify-mermaid-import.py`, and they fail against the previous extractor. The reference documents the three new exit-2 messages.

mysoniw:
I was wrong about the timeline separator, twice, and I should correct that here.

I declined the colon-splitting note on the grounds that Mermaid itself treats every colon as an event separator, and cited mermaid-js/mermaid#6557 and #5699. Those are open renderer issues, not the grammar. The lexer is explicit:

```
timeline.jison:34
  ":"\s(?:[^:\n]|":"(?!\s))+   return 'event';
```

A colon only separates when it is followed by whitespace, and one that is not belongs to the event text. So `2024 : Visit https://example.com/x` is one event, and my `split(":")` was cutting it in half. Fixed — the tokenizer now matches the lexer, and there is a case for it in the verifier.

Reading the .jison sources instead of the rendered docs turned up four more in the same area. `nodeWithoutId` is a real mindmap production, so `((Root))` with no identifier was becoming a rectangle labelled `((Root))`. `accTitle:` / `accDescr:` are valid in every grammar and were producing a phantom gantt task, a phantom timeline period, and a hard failure in quadrant. `#` opens a comment in timeline, per the period rule `[^#:\n]+`. And the shared preprocessing only stripped a comment that starts a line, so a trailing `%% note` cost a mindmap topic its shape and id.

Also rebased onto current main (2.6.5) and bumped to 2.6.6. #131 landed in the same two files while this sat open; its changes are in the flowchart edge lexer and merged cleanly, and I checked line by line that all 96 of them survived.

cathrynlavery (changes requested):
Thanks for the substantial grammar work. Four current-head mismatches need coverage and fixes before merge:

1. Mindmap native edge shapes (`-)Root(-`, `(-Cloud-)`, `))…((`, `)…(`, `((…))`, `{{…}}`, and related forms) are emitted with delimiter text or the wrong shape; multiline Markdown strings inside node delimiters can split into phantom nodes.
2. Timeline permits an event continuation (`: Alpha`) after a standalone period statement; the extractor currently exits 2.
3. Gantt task tags are leading-only. In `id, done, 5d`, the current parser strips the non-leading `done`, loses the explicit id, and misreads the start/duration fields.
4. Quadrant lines with arbitrary suffix text are accepted and silently discard the suffix instead of failing closed.

Please align each case with Mermaid's upstream grammar, add focused positive/negative fixtures, then rerun the full current suite on a fresh rebase and select a unique synchronized manifest version.

mysoniw:
Closing this one. The branch has drifted far enough behind the 2.6.x extractor that rebasing it means rewriting most of the 565 lines against a parser that has moved underneath them, and I'd rather redo it cleanly than keep patching a conflicted branch.

For what it's worth the gap is still there — `mermaid_extract.py` on main still lists gantt, quadrantChart, timeline, and mindmap under `UNSUPPORTED_KINDS`. If you want the grammars I'll open a fresh PR against the current extractor.

PR #126, opened 2026-08-20
fix(commands): quote argument-hint in /profile so the frontmatter parses

xiaoshao9704:
## The defect

`commands/profile.md` line 3, on `main` (`cc2f51f`, v2.5.15):

```yaml
argument-hint: [list|save|load|show|update|reset|delete] [name]
```

A YAML value starting with `[` opens a **flow sequence**. The parser closes that
node at the matching `]`, then finds ` [name]` after it — a second node where
none is allowed:

```
can not read an implicit mapping pair; a colon is missed at line 3, column 64
```

The failure is not limited to that one key. YAML frontmatter is parsed as a
single document, so **the whole block is lost**, including `description` and
`allowed-tools`.

The other three commands are fine, and this is why: their values start with `<`
(`<html-file> [...]`, `<mermaid-file> [...]`), which makes them plain scalars, and
a `[` appearing later in a plain scalar is not special. I scanned all 8 markdown
files with frontmatter in the repo — `commands/profile.md` is the only one that
fails to parse.

## What each host actually does

Measured, not inferred:

| Host | Parser | Behaviour |
|---|---|---|
| Claude Code 2.1.232 | `Bun.YAML.parse` | Throws. Claude catches it, logs `Failed to parse YAML frontmatter …` at **warn** level, and continues with `frontmatter = {}`. The command still loads and runs. |
| Kimi Code 0.36.1 | js-yaml | Throws. `loadPluginCommand` catches and returns `undefined`, so `/profile` **does not exist** there — no error shown. |
| Codex | serde_yaml | Not verified. |

The Claude Code path is the one worth flagging, since that is this plugin's
primary target. `Kp()` in the 2.1.232 bundle:

```js
try { a = YZ(o) }                                   // YZ = Bun.YAML.parse
catch {
  try { a = YZ(o.replace(/^\t+/gm, …)) }            // tab-indentation retry; does not help here
  catch (f) {
    l = f.message
    w(`Failed to parse YAML frontmatter${m}: ${l}`, { level: "warn" })
  }
}
return { frontmatter: a, content: i, … }            // a is still {}
```

So on Claude Code today, `/profile`:

- has no `description` in the slash menu,
- has no `argument-hint`,
- and — the reason I am filing this rather than just patching my own copy —
  **runs without the `allowed-tools` allowlist you declared** (`Read`, `Write`,
  `Edit`, `Bash`, `Glob`). The restriction is silently not applied.

Because the log line is warn-level, nothing surfaces to the user or to CI.

## The fix

Quote the value so YAML reads it as a string instead of a sequence:

```yaml
argument-hint: "[list|save|load|show|update|reset|delete] [name]"
```

After the change all five keys parse, with `allowed-tools` as a proper list:

```json
{
  "description": "Save, load, inspect, update, reset, or delete diagram-design client profiles",
  "argument-hint": "[list|save|load|show|update|reset|delete] [name]",
  "allowed-tools": ["Read", "Write", "Edit", "Bash", "Glob"]
}
```

## Validation

Per `CONTRIBUTING.md`, both manifests are bumped together (2.5.15 → 2.5.16, via
`scripts/bump-plugin-version.py`; it also updated `.factory-plugin`). Gates run
locally:

- `python3 scripts/test-plugin-package.py` — pass
- `python3 scripts/verify-plugin-package.py upstream/main` — pass
- `python3 scripts/verify-docs-sync.py` — pass
- `python3 scripts/verify-semantic-motion.py --markdown-only` — pass
- `claude plugin validate . --strict` — pass

I could not run the full suite locally (this machine has Python 3.9; the repo
asks for 3.10+). The gates above ran clean; the rest are untouched by a
one-line frontmatter change.

Worth noting that `claude plugin validate --strict` passes both before and after
— it validates the marketplace manifest and does not parse command frontmatter,
so it would not have caught this.

## How I found it

Converting the plugin to another agent ecosystem. The converter could not parse
the file, which is what sent me looking at how each host handles it.


xiaoshao9704:
Rebased onto `main` and re-bumped: the manifests are now at 2.5.18, so
`verify-plugin-package.py` passes against the current base.

One thing worth flagging while I was doing it, in case it is useful beyond this
PR. The version gate went stale twice in about ten minutes:

| Landed on `main` | Version | Effect here |
|---|---|---|
| #123 | 2.5.16 | my 2.5.16 was no longer strictly greater |
| #124 | 2.5.17 | my 2.5.17 was no longer strictly greater |

That is not a complaint about either PR — it follows from the two rules
combining: every PR must bump the version, and the version must be strictly
greater than the base. So each merge invalidates the version of every other open
PR at the same instant, and there are 17 of them right now. A contributor can
only win the race by being merged immediately after bumping.

I am happy to keep re-bumping on request, but I would rather not chase it, since
the next merge undoes it again. If it is easier for you to bump this one at merge
time, or to tell me when you are about to merge so I can push a fresh bump,
either works.

The substantive change is unaffected by all of this: still one line in
`commands/profile.md`, quoting `argument-hint` so the frontmatter parses.

cathrynlavery:

cathrynlavery:

cathrynlavery:
@xiaoshao9704, thank you for fixing the /profile argument hint so its YAML frontmatter parses correctly. Compatibility details like this are easy to miss and frustrating for users when they fail. I appreciate you finding it and contributing the focused correction.

PR #105, opened 2026-08-18
docs(high-level): correct effective width checklist

thangldw:
## What does this PR do?

Fixes two contradictions in the High-Level reproducibility checklist: `effective_w` is corrected from `972` to `964` when the 28 px vertical strip and 8 px strip margin are reserved (`1000 - 28 - 8`), matching §§2.1 and 5; the duplicated final item number is corrected from `11` to `13`.

Bumps both plugin manifests from `2.5.1` to `2.5.2` for the package version gate.

## Validation gates

- [x] `python3 scripts/test-lint-a11y.py`
- [x] `python3 scripts/lint-skin.py --all --baseline`
- [x] `python3 scripts/verify-sequence-oauth.py`
- [x] `python3 scripts/verify-drawio-import.py`
- [x] `python3 scripts/verify-mermaid-import.py`
- [x] `git diff --ignore-space-at-eol --exit-code` green after `python3 scripts/build-icons.py`
- [x] Docs updated in the same PR where behavior changed
- [x] Plugin package/version, strict Claude validation, docs sync, semantic motion, geometry, motion, self-check, and treemap gates pass

## Checklist

- [x] No new entries added to `scripts/lint-skin-baseline.txt`
- [x] Generated and source files are consistent (extractor ↔ verifier ↔ reference ↔ command)
- [x] Accessible SVG contract satisfied for new/changed examples (no examples changed)
- [x] Code of Conduct respected

## Related issues

No issue filed; this is a standalone documentation consistency fix.

PR #129, opened 2026-08-20
feat(bar): add histogram as a Bar variant

ggplab:
## What

Histogram as a **Bar variant**: one continuous variable cut into equal contiguous bins, height = count. Ships the slopegraph-shaped set — variant spec in `type-bar.md`, light/dark/full examples, a gate (`verify-histogram.py`), its adversarial suite (27 cases), CI wiring, and one gallery tab (`Bar · histogram`, eyebrow 31 — filling the existing gap between 30 and 32).

## Why this is a routing gap

- The selection table has no row for **one variable's distribution**. Scatter's row starts at *two* variables, and its overplotting advice ("more → bin into a density contour") is binning applied to a two-variable relationship plot — binning **one** variable answers a different question (shape, not relationship) and had nowhere to route.
- Histogram is not in ADR 0007's rejected list, and it introduces **no new layout grammar** — it is bar's length encoding with a tighter contract (contiguous equal bins, edge-labelled boundaries, zero-anchored counts). So: a variant, the dumbbell/slopegraph path. The 38-type count is untouched.

## The contract, executable

Every histogram lie renders perfectly, so the reference's rules are gated by `scripts/verify-histogram.py --all`:

1. equal bin widths, declared **and** drawn (widening one bin raises its bar with no data moving — the signature lie)
2. bins tile: bin k ends where k+1 starts, in values and pixels (gap 0 is the variant's grammar)
3. one shared baseline, the printed `0` tick on it, one linear px-per-count scale across bars **and** ticks
4. printed **n equals the sum of the bins** (dropped outliers are usually the story)
5. an optional marker sits where the scale puts its declared value — and a marker claiming to be the *mean* must lie within half a bin of the binned mean (no placement by eye)
6. transforms rejected on verified geometry, bound labels, and groups
7. **fail closed**: a file presenting as a histogram with nothing parseable is a finding, and `--all` asserts the three shipped examples stay in scope — a scope heuristic that stops matching them cannot green the gate by blinding it (the fail-open class #127 closed for treemap)

The bin budget (6–12) and marker weights (ink 55%/40%) are **parsed out of `type-bar.md` at check time** — the `verify-dumbbell.py` coupling — so prose and thresholds cannot drift apart; an unparseable reference is itself a finding. `scripts/test-verify-histogram.py` holds both polarities: every mutation above is caught, the shipped bar example stays out of scope, an empty bin drawn at height 0 stays clean, and fixtures live in a per-process temp dir.

## Documented departures from the parent chart

- **6–12 bins vs the 4–8 bar cap** — the cap exists because categorical bars are read one at a time; a histogram's silhouette is read as one shape, and a shape needs more segments than a comparison needs bars (the slopegraph precedent of arguing its own budget from the reading, `type-line.md` series count).
- **All paper masks painted before any bar body** — with shared edges, a neighbour's opaque mask painted later erases the outer half of the shared 1px stroke. Interleaving is safe only with gaps.
- **Bin edges exempt from the 4px grid** — data coordinates, the dumbbell's reason: snapping an edge moves a bin boundary. (At 10 bins over 880px the shipped example happens to comply.)
- Focal stays 1-max (the modal bin here); one reference marker max; edge ticks may print every second edge when numerals collide, so the gate checks every printed edge without requiring all of them.

## Verified

Full local suite green against `upstream/main` as base: `test-plugin-package`, `verify-plugin-package upstream/main`, `plugin validate --strict`, `test-lint-a11y`, `verify-semantic-motion` (both modes), `verify-motion --shipped`, `lint-skin --all --baseline` (141 checked, 0 findings), `verify-sequence-oauth`, `verify-docs-sync` + tests, `verify-screenshot-freshness`, `test-self-check`, `verify-geometry --all` + tests, `verify-treemap --all` + tests, `verify-dumbbell` + tests, `verify-slopegraph --all` + tests, `verify-histogram --all` + tests, icon build round-trip clean. Light and dark renders eyeballed in headless Chrome — shared edges draw at full 1px, dark hairlines inverted.

SKILL.md edit is one routing-row phrase (+~60 bytes, 39,240/40,000 against the ADR 0004 cap — no trimming owed). No screenshots owed: variants add no `type-*.md` row, `verify-screenshot-freshness.py` confirmed clean.

## Version & open-PR relations

- **2.5.21** (patch, via `bump-plugin-version.py`, all three manifests) — unique across open PRs (#126/#114 hold 2.5.18, #113 2.5.17, #112 2.5.16, #105 2.5.7, #101 2.6.0). If another versioned PR merges first, I will rebase and re-bump.
- **#114** (SKILL.md byte headroom): this PR touches one SKILL.md row; trivial to merge in either order, happy to rebase.
- **#112** (gate the type count): no interaction — this PR adds no type and moves no counter.


philyuchkoff (approved):
# Review: feat(bar) histogram as a Bar variant — PR #129

@ggplab

Excellent shape — this is by far the strongest variant-gate PR I've reviewed in this repo. The checker (verify-histogram.py) is genuinely rigorous: it catches unequal bins (the histogram's signature lie), tiling gaps, floating bars, bent scales, n mismatches, label honesty, marker accuracy, and transforms — all with the right fail-closed and fail-positive guardrails. The adversarial test suite follows the exact pattern of existing verifier tests (slopegraph, sankey, treemap) and ships 27 cases, both polarities.

Specifics:

1. Bin contiguity as a hard invariant (check_bins) — "a gap is what says categories" — this distinguishes bar chart from histogram not as a visual choice but as a grammar rule. Correct.

2. Fail-closed design (looks_like_histogram → parse_bins fails if < 3 parseable bins) — this is exactly the same class of bug #127 closed for treemap. Good hygiene.

3. Reference coupling (reference_tokens) — the bin-count budget and marker weights are parsed from type-bar.md, not hardcoded. If the prose changes the budget, the checker changes with it. Same pattern as verify-dumbbell.py.

4. Zero-bin adversarial test is subtle and correct — data-count=0 with height 0 and its label still passes. A 0-height bar is the only correct render of an empty bin. Avoiding a false positive is important.

One minor note: in looks_like_histogram, the DECLARES_RE.search would match files  bar\uttindires(later \ bin-star fields — this is the right conservative behavior (fail rather than silently pass) ) but it would fire on a non-histogram bar chart. The comment already addresses this — "deliberately generous" — so fine.

**Version bump**: 2.6.1 → 2.6.3 correct.

Gates look clean — shipped examples pass, all checks green.

LGTM, approved.

cathrynlavery (changes requested):
Two verifier gaps need to be fixed before rebasing:

- Repeated `n` metadata is currently overwritten, so contradictory sample counts can pass. Require exactly one declaration or validate every occurrence and reject disagreement.
- The CSS transform check scans every declaration in the document and rejects transforms on unrelated selectors. Scope it to selectors/elements that can reach verified histogram geometry, with positive and negative fixtures.

Please add adversarial tests for both, then rebase and resolve conflicts, choose a unique synchronized version, and run the complete current suite including the later ridgeline and bubble gates.

ggplab:
@cathrynlavery Both gaps fixed in b853f3d, on top of a fresh rebase.

**Repeated `n`** — went with your second option: every `data-role="n"` declaration is validated individually (printed vs declared, and against the bins' sum), and a set of declarations that disagree is its own finding. Single-slot shadowing is gone; the commit message records the exact failure it closes. Fixtures: disagreeing duplicates fire two findings (disagreement + the wrong one vs the sum), agreeing duplicates stay clean.

**CSS transform scope** — the check now asks reachability instead of scanning the document: a transform-carrying rule is a finding when its selector names an SVG element type, the universal or an attribute selector, or a class/id the SVG markup actually uses; a rule scoped to page chrome (e.g. a card hover) is allowed. A transform the parser cannot attribute to any rule is still a finding — unparseable style is never a pass. Fixtures: `.frame:hover` passes; `text { … }`, `[data-bin-start] { … }`, and `.bin { … }` with `class="bin"` on a bar all fire.

Seven new cases, 34 total, both polarities.

Housekeeping per your list: rebased onto `648c2a5`, version **2.6.8** (2.6.5 is main; 2.6.6/2.6.7/2.6.9 are claimed by open PRs), and the complete current suite is green locally against `upstream/main` — including `verify-ridgeline --all`, `verify-bubble --all`, `verify-polar`, their checkers, and `lint-render` on the three histogram examples (0 findings).

cathrynlavery (changes requested):
Thanks for the fast turnaround on both gaps — the per-declaration `n` validation and the reachability-based CSS scope are exactly what I asked for, and the new fixtures hold both polarities. I re-ran the suite at b853f3d locally: `verify-histogram.py --all`, `test-verify-histogram.py` (34 cases), `lint-skin.py --all --baseline` (150 files, 0 findings), `verify-geometry.py --all`, and docs-sync are all green, and the CI rollup is green across the matrix.

Two things before this lands:

1. **The reachability model stops one level too early.** `check_css_transforms` inventories classes/ids only *inside* the `<svg>` span, but the shipped examples wrap the SVG in `<div class="frame">`, and an ancestor's transform reaches every verified mark through inheritance. I reproduced it: adding `.frame { transform: scaleY(0.55) skewX(12deg); }` to a copy of `example-histogram.html` distorts every bar and the count scale, and `verify-histogram.py` passes it with exit 0. Please treat selectors that can match an ancestor of the SVG as reachable (inventory classes/ids on the SVG's ancestor chain, not just inside it). The current `.frame:hover` fixture then needs a rethink: if you want to keep a carve-out for interactive lifts, make it `:hover`/`:focus`-only and add both fixtures — a *static* ancestor transform must be a finding, a hover-only lift may pass.

2. **Rebase over #160 and re-bump.** Main is now 2.6.11, so 2.6.8 no longer clears the version gate (I get "must increase relative to origin/main: 2.6.11 -> 2.6.8" on all three manifests), and the conflicts are the usual mechanical spots: the three manifests, `ci.yml`, `CONTRIBUTING.md`, `SKILL.md`, and the gallery index. After rebasing, please run the full current gate table — #160 added `verify-skin-polarity.py --all`, `test-maintainer-policy.py`, and the beeswarm/bump verifier suites, which postdate your base.

For what it's worth, the variant is wanted: the ridgeline and beeswarm references that just landed both route "a single distribution" to a histogram by name, so this fills a hole main's own prose now points at. The checker's core — bin tiling, the zero-anchored shared scale, n-equals-sum, reference-coupled budgets, fail-closed scope — is in excellent shape.

cathrynlavery (changes requested):
The histogram data and visuals are sound, but browser-valid cases evade the verifier: static ancestor transforms, duplicate attributes with browser first-wins behavior, quoted > delimiters, and non-data CSS attribute selectors. Use quote-aware, first-wins parsing and complete selector/ancestor reachability with adversarial fixtures. Fix the nested gallery button, rebase onto current main, drop manifests, register against all current gates, attach visual proof, and refresh review/checks.

cathrynlavery:
Thanks for carrying the earlier verifier fixes through the current rebase; the quote-aware tag parsing, browser first-wins attributes, and ancestor class/id inventory are all present on head `52bea3d`.

One browser-valid transform path is still outside the gate: inline CSS. `check_transforms()` only checks a literal `transform` attribute on `<svg>/<g>` and data-bound SVG elements, while `check_css_transforms()` only inspects `<style>` blocks. A static ancestor such as:

```html
<div class="frame" style="transform:scaleY(.55)">
  <svg>…</svg>
</div>
```

therefore moves every verified bar and axis in the browser, but neither path examines the `style` attribute. The same bypass works directly on a bound `<rect style="transform:…">`.

Please reject non-interactive inline `transform` declarations on the SVG ancestor chain and on verified SVG geometry/labels, using the same first-wins attributes produced by `HTMLParser`. Focused adversarial fixtures for both the ancestor and bound-rect forms would close this without widening the rule to unrelated page chrome. Runtime execution remains unverified in this scheduled review; this finding is from static inspection of `scripts/verify-histogram.py` on the exact head.

PR #131, opened 2026-08-20
fix(mermaid): parse compact multidirectional labels

### LINKED ISSUE #80: mermaid_extract mis-parses compact edge-label syntax (B--yes-->C): phantom nodes, labels dropped

cmondragon023:
## Summary

- retain left arrow, circle, and cross markers when parsing compact labeled Mermaid links
- parse `A<--yes-->B`, `A o--yes--o B`, and `A x--blocked--x B` without phantom nodes or malformed-edge failures
- avoid consuming source identifiers ending in marker-like `x` or `o` characters
- preserve single-character source IDs in `x--yes-->B` and `o--go-->C`
- preserve `x`/`o` endpoints in compact, spaced, and pipe-labeled edge chains
- add adversarial regression coverage for labels, endpoints, bidirectional flags, and arrowhead semantics
- merge current `main` and bump the synchronized Claude, Codex, and Factory manifests to the collision-free version `2.6.4`

Closes #80

## Testing

- all 31 local commands in `.maintainer-policy.json` passed on exact head `1d04107`
- `python3 scripts/verify-mermaid-import.py`
- `python3 scripts/test-plugin-package.py`
- `python3 scripts/verify-plugin-package.py origin/main`
- `npx --yes @anthropic-ai/claude-code@2.1.229 plugin validate . --strict`
- `git diff --check`

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, polar chart (polar/radial lollipop), loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, DP security matrix, sankey, fishbone, Wardley map, kanban, user journey, deployment, dependency graph, UML class, story map, or database schema diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.6.1",
+  "version": "2.6.2",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, polar chart (polar/radial lollipop), loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, DP security matrix, sankey, fishbone, Wardley map, kanban, user journey, deployment, dependency graph, UML class, story map, or database schema diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.6.2",
+  "version": "2.6.4",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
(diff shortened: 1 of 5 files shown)


cmondragon023:
Rebased onto current `main` and addressed the marker ambiguity: compact labels no longer consume a source ID ending in `x` or `o` (including `Box--yes-->C`). Added regression coverage for both endings, kept the upstream plugin metadata, and bumped the synchronized manifests to 2.6.2. Mermaid/package/docs/self-check gates pass.

philyuchkoff (commented):
# Review: fix(mermaid) compact multidirectional labels — PR #131

@cmondragon023 

Nice fix — the Mermaid spec really does define these compact labeled links and the current extractor was silently dropping left-side markers, discarding labels, or (worse) turning `o--yes--o` into two zero-width edges. The addition of named regex groups plus the stricter non-backtracking opening alternative is the right level of surgery. Specific observations:

1. The regex change (`<(?:--|-\.|==)`) correctly anchors the opening token to a `<` prefix **or** a standalone mid-line `o`/`x`. The negated-lookbehind `(?<![\w.:-])` prevents false matches on node IDs ending in `x` or `o` (like `Box` or `Echo` — and the test fixtures exercise this explicitly). 

2. The `_operator_style` function already handled `token.startswith('x')` and `token.startswith('o')` for Undirected links (unlabeled). With this PR, compact labeled links that carry left-side markers (`A<--yes-->B`) also flow through that function — the combined token correctly flags `bidirectional` as `True` for the `<-- -->` form. The non-compact `<== retry ==>` form already worked; the gap was only in the whitespace-less compact path.

3. One subtlety worth calling out: the negated-lookbehind `(?<![\w.:-])` excludes `-` from the valid prefix characters. That means `Node-1 o--yes--o Node-2` (a node name ending in dash) would still fire the `o` — but I can't think of a Mermaid node naming convention where a leading `<|o|x>` directly follows a literal `-` in a sensible way, so this is fine in practice. If this ever becomes a false positive, the lookbehind class could be expanded to include `-`.

4. The tests in `verify-mermaid-import.py` are thorough — juniors `Box` and `Echo` exercise the negated-lookbehind, the compact-label slice assertion at `[9]` and `[10]` catches actual phantom-node regressions, and the bidirectional-arrowhead triple-check covers all three marker types. 

5. version bump: manifests now declare `2.6.2`, but the PR body says `2.5.21` — the actual bump is correct (2.6.1→2.6.2), the body is stale. Not a blocker, but it would be nice to update the description for the audit log.

Minor: the comment on line 605–610 still documents the old behavior (“the operator characters themselves may not open one”). The PR adds a new comment inside the diff that updates this — but only inside the diff view of the PR. The actual repo comment is being replaced by the new text in the PR that adds `A o--yes--o B`. Wait, I see that the PR adds the documentation comment alongside the regex update — correct. The old comment was part of the replaced lines.

**One more thing**: the `--` and `-.` and `==` as opening tokens no longer have a negated-lookbehind guard (`(?<![\w.:-])...` is only on the `o`/`x` alternative). This means `Box--yes-->C` could still be problematic if `--yes-->` is mistokenized as part of a label. But this is the existing behavior (the `--` is `<opening>` which has always been permissive), and the test fixture confirms it's fine.

**Summary**: scope is correct, tests cover all changed code paths, docs comment updated, inline with comment. One doc nit on the PR body — bump version is 2.6.1→2.6.2, not 2.5.21. Non-blocking. LGTM on the code side.

Happy to approve on your next pipeline run.

PR #138, opened 2026-08-21
fix(assets): apply the dark skin to example-line-dark

0xDarkMatter:
example-line-dark.html shipped with the light palette — paper `#f5f5f5` and a light background rect — so it renders as a second light variant. Flipped the tokens to the dark set every other `-dark` example uses (paper `#2d3142`, ink `#f5f5f5`, muted `#bfc0c0`, accent `#f08a59`).

lint-skin passes either way since it doesn't compare skin to filename. `lint-skin --all --baseline`, `verify-geometry --all`, `self_check` on the file, and the package gate are all green locally. Patch bump to 2.6.2.


cathrynlavery (changes requested):
The actual dark-skin correction is accepted: I rendered exact head `43b1f2c`, and the line example now matches the dark palette without unnecessary changes. The remaining blockers are branch drift only. Please rebase on current `main`, resolve conflicts, bump all three manifests to a unique increasing version, and run the complete current gate suite (including the ridgeline and bubble verifier/test pairs added since this branch).

cathrynlavery:
Quick status update after #160 landed: this fix is still needed and still right. Current main ships `example-line-dark.html` with the light tokens (`--color-paper:#f5f5f5`, background rect `#f5f5f5`), and your flip to the standard dark set matches every other `-dark` example — I re-verified the file on main today. What changed since the last review is that #160 moved the manifests to 2.6.11, so this branch now conflicts and the package gate fails (2.6.7 no longer increases over main). To land it: rebase onto current main, run `scripts/bump-plugin-version.py` for the next free patch version, and run the current gate suite — #160 added `verify-skin-polarity.py --all`, `test-maintainer-policy.py`, and the bump/beeswarm verifier pairs since your branch. Everything else is already green on your head as-is (lint-skin, verify-geometry, verify-docs-sync, the full CI matrix), so this should be a quick relanding.

cathrynlavery:
Thank you for catching this. The accepted Line dark-skin correction is on main through #174 with canonical colors and a regression check. #174 credits the original patch; only obsolete manifest bumps remain here, and the branch conflicts with main. Closing as superseded.

PR #75, opened 2026-08-16
fix(lint): allow CJK fallback stacks documented in output-spec.md

atirna:
### Problem

`output-spec.md` §4 documents CJK fallback stacks for non-Latin labels, but `lint-skin.py` rejects the same families as unsupported. Following the spec fails CI.

Repro on `main`:

```bash
python3 scripts/lint-skin.py cjk-font-probe.html
# font-family: unsupported font family: Hiragino Sans, Noto Sans JP, Yu Gothic
# font-family: unsupported font family: Noto Sans Mono CJK JP
```

### Fix

Add the four families named in `output-spec.md` to the linter allowlist:

- `Hiragino Sans`, `Noto Sans JP`, `Yu Gothic` (sans stack)
- `Noto Sans Mono CJK JP` (mono stack)

These stay per-element fallbacks, not skin roles, so I kept them in the linter rather than the style-guide Typography table.

Only the documented families pass. Sibling variants like `Yu Gothic UI` or `Noto Sans Mono CJK KR` are still rejected (tested).

### Verification

- before, on `main`: issue #74 probe → 2 font-family findings, exit 1
- after: same probe → 0 findings, exit 0
- `python3 scripts/test-lint-a11y.py` (new pass + reject cases for each stack)
- `python3 scripts/lint-skin.py --all --baseline` → 102 files, 0 findings
- `python3 scripts/verify-plugin-package.py origin/main`

Fixes #74

Made with [Cursor](https://cursor.com)

PR #145, opened 2026-08-22
fix(factory): sync maintainer version policy

HsienW:
## What does this PR do?

Factory Droid is already part of the native plugin package and is covered by the package verifier and version bump helper, but two maintainer-facing surfaces were still describing the older two-manifest setup.

This PR keeps those contracts in sync by:

- adding Factory Droid to the supported-host wording in `CONTRIBUTING.md`
- replacing host-count-specific wording with native-host-neutral wording
- adding `.factory-plugin/plugin.json` to the synchronized manifest list in `.maintainer-policy.json`
- bumping the three native plugin manifests together to `2.6.6` using the repository helper

No plugin runtime, install, or host behavior changes here — this just brings the contributor guidance and machine-readable maintainer policy in line with the Factory Droid support that already exists.

## Visual proof

Not applicable — no diagram or visual output changes.

## Validation gates

- [x] `python3 scripts/test-lint-a11y.py`
- [x] `python3 scripts/lint-skin.py --all --baseline`
- [x] `python3 scripts/verify-sequence-oauth.py`
- [x] `python3 scripts/verify-drawio-import.py`
- [x] `python3 scripts/verify-mermaid-import.py`
- [ ] `git diff --exit-code` green after `python3 scripts/build-icons.py` (if icons changed) — N/A, no icon changes
- [ ] Docs updated in the same PR where behavior changed — N/A, no runtime behavior change

Additional checks relevant to this change:

- [x] `python scripts/test-plugin-package.py`
- [x] `python scripts/verify-plugin-package.py upstream/main`
- [x] `claude plugin validate . --strict`
- [x] `python scripts/verify-docs-sync.py`
- [x] `python scripts/test-verify-docs-sync.py`
- [x] `git diff --check upstream/main...HEAD`

I also ran the full local validation suite from `CONTRIBUTING.md`. A few unrelated baseline/environment checks fail the same way on a clean `upstream/main` checkout:

- `verify-semantic-motion.py --markdown-only` — `SKILL.md` is currently over the 40,000-byte limit
- `lint-render.py --self-test` / `--all` — Playwright is not installed in my local Windows environment
- `verify-doctor.py` — reproduces a Windows `WinError 1312`
- `verify-screenshot-freshness.py` — existing canonical screenshots are stale against their current sources
- `test-verify-treemap.py` — reproduces a Windows CP950 `UnicodeEncodeError`

I left those unrelated failures untouched to keep this PR scoped to the Factory maintainer-policy sync.

## Checklist

[...]

PR #146, opened 2026-08-22
fix(lint): allow the Chinese fallback stacks output-spec.md already documents

lfsinzzu-jpg:
## Problem

`output-spec.md` instructs authors to extend the font family "when labels contain Japanese, **Chinese**, or Korean text" — but it only ships Japanese and Korean example stacks, and `ALLOWED_FONTS` in `scripts/lint-skin.py` contains no Chinese families at all.

The result is that a literal `font-family` carrying a Chinese stack is rejected, while the equivalent Japanese stack passes:

```
example-venn.html:21: font-family: unsupported font family: PingFang SC, Noto Sans SC, Microsoft YaHei
```

Found via a controlled comparison on a copy of `example-venn.html`, varying only the font stack:

| Arm | Stack | Declaration | Result |
|---|---|---|---|
| Japanese | `Hiragino Sans, Noto Sans JP, Yu Gothic` | literal | PASS |
| **Chinese** | `PingFang SC, Noto Sans SC, Microsoft YaHei` | literal | **FAIL** |
| Comic Sans MS (negative control) | — | literal | FAIL (checker is live) |

The negative-control arm is included because the same three stacks all pass when declared through a `--font-*` CSS custom property — `FONT_CSS_RE` only matches literal `font-family:` declarations, so a variable-only test would have shown a false green.

## Change

Mirrors the Japanese (#75) and Korean (#83) fixes:

- **`scripts/lint-skin.py`** — add Simplified (`PingFang SC`, `Noto Sans SC`, `Microsoft YaHei`, `Noto Sans Mono CJK SC`) and Traditional (`PingFang TC`, `Noto Sans TC`, `Microsoft JhengHei`, `Noto Sans Mono CJK TC`) families, following the existing Apple-native / Noto / Microsoft-native pattern.
- **`skills/diagram-design/references/output-spec.md`** — add the SC and TC example stacks and the mono variants; document why a Japanese stack cannot stand in for Chinese (subset coverage, plus Japanese glyph variants for Simplified forms) and why SC and TC are separate stacks.
- **`scripts/test-lint-a11y.py`** — add `zh-name-stack`, `zh-mono-stack`, `zh-css-stack`, `zh-tc-stack`, matching the existing `kr-*` cases.
- Version bumped 2.6.5 → 2.6.6 via `scripts/bump-plugin-version.py`.

## Verification

The new cases were confirmed to **fail without** the `ALLOWED_FONTS` change:

```
AssertionError: zh-name-stack: expected exit 0, got 1
  zh-name-stack.html:6: font-family: unsupported font family: PingFang SC, Noto Sans SC, Microsoft YaHei
```

and pass with it. Gates run locally on Windows / Python 3.12, all green:

`test-lint-a11y.py` · `lint-skin.py --all --baseline` · `test-plugin-package.py` · `verify-plugin-package.py` · `verify-docs-sync.py` · `test-verify-docs-sync.py` · `test-fix-mojibake.py` · `test-verify-geometry.py` · `test-verify-treemap.py`

Scope is deliberately limited to the lint allowlist and its documentation — no rendering or geometry behaviour is touched. Happy to split the docs change out if you would rather keep the PR to the allowlist alone.


cathrynlavery (changes requested):
Thanks @lfsinzzu-jpg — this is a well-evidenced fix, and the controlled comparison (including the negative-control arm) made it easy to trust. I confirmed on current `main` that `ALLOWED_FONTS` in `scripts/lint-skin.py` still has no Chinese families while `output-spec.md` tells authors to extend the stack for "Japanese, Chinese, or Korean" labels, so the gap is real. The additions follow the existing exact-token Apple/Noto/Microsoft pattern (the linter stays strict), and the `zh-*` test cases mirroring the `kr-*` ones are exactly what we want.

The branch is conflicting, so one rebase pass is needed:

- `main` moved under you: `noto serif kr` was added next to your insertion point in `ALLOWED_FONTS`, and the "Non-Latin labels" section of `output-spec.md` was reworked (the Korean example stack now leads with `Noto Sans KR` and links to `style-guide.md#korean-labels`). Please re-apply the SC/TC additions against the current text rather than the old paragraph.
- Re-run `python3 scripts/bump-plugin-version.py` after rebasing — main is at 2.6.11, so 2.6.6 no longer passes `verify-plugin-package.py origin/main` (and it collides with #145's bump; versions must be unique across open PRs).

Everything else checks out at your current head — `test-lint-a11y.py`, `lint-skin.py --all --baseline`, and the package tests all pass locally for me, and CI is green across the matrix — so with the rebase and a fresh bump this looks mergeable.

cathrynlavery:
Thanks for the controlled comparison and regression coverage. The accepted Chinese allowlist, documentation, and four lint fixtures landed on main through #174, preserving LiuQQ as author of commit 62234ad; only the stale manifest bump was omitted. Closing as superseded. #175 remains separate follow-up for TC webfonts/rules.

PR #39, opened 2026-08-12
feat: shared SVG extraction tooling for Figma export (branch-independent slice)

cathrynlavery:
## Context

First slice of the [send-to-Figma plan](docs/plans/2026-08-12-001-feat-figma-code-to-canvas-export-plan.md) (eng-review cleared 2026-08-12). The plan's Phase 0 spike (U1) is a hard decision gate that needs live Figma MCP OAuth in both Claude Code and Codex plus a Full-seat account — not yet configured — so this PR ships only the work that is required in **both** spike outcomes. The Code to Canvas machinery (`build`/`serve`, `references/figma.md`, `commands/figma.md`, `prompts/figma.md`, README/manifest packaging) stays deferred until the spike runs and the branch decision is written into the plan.

## What shipped

- **`skills/diagram-design/scripts/figma_capture.py`** — stdlib-only `extract` + `check` subcommands (plan U2 partial, decisions 4A/T1/6A). Balanced-tag `html.parser` isolation preserves the nine shipped nested-SVG examples that first-`</svg>` regex extraction truncates. `check` validates viewBox (4 finite numbers, positive dimensions), refuses galleries, `<script>`, event handlers, `javascript:` URLs, and any external `url()`/`@import`/`href` outside the two Google Fonts hosts; `foreignObject` warns (the medallion examples are legitimate) while active content inside it still rejects. Extract output is standalone, well-formed XML — the fonts `@import` is injected with `&amp;` escaping and verified with `ET.fromstring` before writing. Sources are never modified (R3, hash-proven).
- **`skills/diagram-design/references/export.md`** (plan U5, decision 5A) — the SVG procedure now delegates to `extract` with the prose kept as the no-script fallback; fixes the stale `commands/export.md` pointer; corrects a latent XML bug in the doc's own `@import` snippet (raw `&` is a fatal parse error in a standalone SVG).
- **`scripts/verify-figma-export.py` + `scripts/fixtures/sample-nested-svg.html` + CI step** (plan U3 partial, decision 7A) — 12 case families pinning every shipped behavior, including the CRITICAL extract-vs-export.md parity regression guard and the repo-wide reference→command pointer-resolution check. Wired into the existing 3-OS × 2-Python matrix with a summary-table row.

## Test evidence

- `verify-figma-export.py`: 12/12 green locally; deliberate mutations fail it (stale pointer → case 12; re-introduced first-`</svg>` truncation → case 4).
[...]

PR #151, opened 2026-08-23
fix(export): normalize rgba()/transparent for PowerPoint's SVG importer

### LINKED ISSUE #148: [Bug]: SVG export is unusable in PowerPoint — rgba() and transparent render as opaque black
### Area

Export (SVG / PNG)

### What happened?

`references/export.md` says the SVG export is "diagram-only — just the `<svg>` node… suitable for Figma, slides, social cards, or blog images."

For **slides it is not**, if the slide tool is PowerPoint. PowerPoint's SVG importer is SVG 1.1 and does not understand CSS colour notation. It renders `rgba(...)` and `transparent` as **opaque black**, so a barely-there tint becomes a solid dark box that swallows the label inside it.

The export procedure normalises `xmlns`, `viewBox`, and the font `@import` (including the `&amp;` XML-escaping trap, nicely caught), but it does not touch colours. So an SVG produced by following the documented steps is well-formed XML, opens correctly in a browser, and is visibly wrong on a slide — with no error at any stage.

Scope, measured on the 2.6.5 example assets:

| Notation | Occurrences | Files affected |
| --- | --- | --- |
| `rgba(...)` | 2850 | 149 / 149 |
| `transparent` | 277 | 49 / 149 |

No `hsl()`, `color-mix()`, or `oklch()` — so `rgba()` and `transparent` are the whole class.

Concretely, in `assets/example-architecture.html`:

```svg
<rect x="220" y="240" width="144" height="64" rx="6"
      fill="rgba(45,49,66,0.03)" stroke="rgba(45,49,66,0.30)" stroke-width="1"/>
```

In a browser that is a near-white card with a hairline border. In PowerPoint it is a solid navy block, and the "Cloudflare" label rendered on top of it disappears.

The routing table in `output-spec.md` sends PowerPoint users to `png @2`, which sidesteps this — but that reads as a quality preference, not as "SVG is broken here". Anyone who wants editable vector on a slide follows the SVG path and gets a silently wrong diagram.

### Steps to reproduce

1. Follow the `references/export.md` SVG procedure on `assets/example-architecture.html` to produce `example-architecture.svg`.
2. Insert that `.svg` into a PowerPoint slide (Insert > Pictures). It lands as a native graphic, shape type 28.
3. Insert the PNG export of the same diagram on a second slide as a control.
4. Render both slides to images and compare.

Result: the SVG slide has solid dark fills where the browser shows light tinted cards, and the labels inside those shapes are gone. The PNG slide is correct.

Replacing every `rgba()`/`transparent` occurrence as below and repeating the test makes the two slides match.

### Environment

Windows 11, PowerPoint 16.0 (365), diagram-design 2.6.5, Claude Code

### Suggested fix

Add a colour-normalisation step to the SVG export procedure. It is lossless — browsers render both forms identically, so nothing regresses for the Figma/Illustrator/browser destinations.

```
fill="rgba(r,g,b,a)"    ->  fill="#rrggbb" fill-opacity="a"
stroke="rgba(r,g,b,a)"  ->  stroke="#rrggbb" stroke-opacity="a"
fill|stroke="transparent"  ->  fill|stroke="none"
```

Two attribute-level regexes cover all 3127 occurrences. Happy to open a PR against `references/export.md` if that shape of fix looks right to you.

Worth a line in the export caveats either way: the existing note warns that offline Illustrator and some Figma paths substitute *typography*. The colour failure is more severe, because typography substitution still reads as the same diagram and this does not.

Jenil-1714:
Fixes #148.

## Problem

`references/export.md`'s SVG export procedure produces a well-formed, browser-correct standalone SVG, but doesn't normalize color notation. This design system's tokens are authored as `rgba(...)` (see `style-guide.md`), which every consumer named in the export doc — browsers, Figma, Illustrator — renders correctly.

PowerPoint's SVG importer is SVG 1.1 and doesn't understand CSS Color Module notation: it renders `rgba(...)` and `transparent` as **opaque black**. A barely-there tinted card becomes a solid dark block, and any label drawn on top of it disappears — silently, with no error at export or import time.

Verified current scope on `main` (matches the issue's numbers exactly): 2850 `rgba()` occurrences across all 149 example files, `transparent` in 49 of them, both used exclusively as `fill=`/`stroke=` presentation-attribute values (never inside a `style="..."` attribute or a `<style>` block).

## Fix

Add a color-normalization step to the SVG export procedure (step 4, between the existing standalone-ization bullets and the XML-declaration step):

- `fill="rgba(r,g,b,a)"` → `fill="#rrggbb"` plus a new `fill-opacity="a"` attribute (same for `stroke`/`stroke-opacity`).
- `fill="transparent"` → `fill="none"` (same for `stroke`).

This is lossless — every renderer that already displayed the diagram correctly renders the transformed attributes identically, since `fill-opacity`/`stroke-opacity` and `none` are core SVG 1.1 presentation attributes. The source HTML and design tokens are untouched; only the extracted, exported SVG is normalized. This mirrors how the same procedure already XML-escapes the font `@import` ampersands (#38) rather than touching the source.

Doc-only change — the SVG export path has no shipped script (unlike PNG export, which does), so there's nothing else to update.

## Testing

- Wrote the documented regex transform standalone and ran it against the full extracted `<svg>` from `example-architecture.html`: all 22 `rgba()` and 5 `transparent` occurrences convert, output is well-formed, and a spot-checked value (`rgba(45,49,66,0.03)` → `#2d3142` + `fill-opacity="0.03"`) matches the issue's suggested transform exactly.
- Ran the full CONTRIBUTING.md gate suite (`verify-docs-sync`, `test-verify-docs-sync`, `lint-skin --all --baseline`, `test-plugin-package`, `verify-plugin-package`, `claude plugin validate --strict` via npx, `test-lint-a11y`, `verify-semantic-motion` ×2, `test-verify-motion`, `verify-doctor`/`test-verify-doctor`, `verify-sequence-oauth`, `verify-drawio-import`/`test-verify-drawio-import`, `verify-mermaid-import`, `verify-screenshot-freshness`, `test-self-check`, `verify-geometry --all`/`test-verify-geometry`, `verify-treemap --all`/`test-verify-treemap`, `verify-polar`/`test-verify-polar`, `verify-dumbbell`/`test-verify-dumbbell`, `verify-slopegraph --all`/`test-verify-slopegraph`, `verify-ridgeline --all`/`test-verify-ridgeline`, `verify-sankey --all`/`test-verify-sankey`, `verify-bubble --all`/`test-verify-bubble`) — all pass.
- `lint-render.py` (and the Playwright-backed PNG export path) couldn't run — Playwright/Chromium isn't installed in this environment. This change doesn't touch any HTML/JS/rendering surface, so I don't expect it to be affected, but flagging honestly rather than claiming a gate I didn't run.
- Bumped the plugin version (`scripts/bump-plugin-version.py`, patch) per CONTRIBUTING.md; `verify-plugin-package.py origin/main` passes.
- Confirmed `python3 scripts/build-icons.py` produces an unrelated regeneration diff on a clean, unmodified `main` in this environment (icon-fetch nondeterminism, not caused by this change) — left untouched.

## Not in scope

`output-spec.md` still routes PowerPoint/Keynote/Slides to PNG by default, which stays reasonable general guidance. This PR only fixes the SVG path so it's correct if someone follows it for editable-vector needs, per the issue.


cathrynlavery:
Thanks for isolating and proving the PowerPoint color issue. I’m going to carry this into a clean current-main export integration PR together with the primary `figma_capture.py` extractor from draft #39, because normalizing only the retiring manual snippet would be regressed by that change. I’ll preserve your authorship with a `Co-authored-by` trailer and link the integration PR here. We’ll leave this PR open until the integrated replacement lands, then close it as superseded—not rejected.

cathrynlavery (approved):
Rechecked exact head 12f2a5e4a70686927c988e5778d2f9acb83105ea against current origin/main 2724fd2efd8c6737f6fa704fbf5da52d67375497. The complete diff, commits, discussion, timeline, and resolved review thread are clean; all 54 declared policy gates and all 8 hosted checks pass; ADR 0009 is satisfied; and this documentation-only change correctly normalizes the repository’s shipped SVG presentation-attribute forms for strict SVG 1.1 consumers. Approved for squash merge.

PR #153, opened 2026-08-24
fix(mermaid): retain quoted labeled-link text

Jenil-1714:
## What does this PR do?

Quoted text on a spaced labeled link — `A-- "text" -->B` — was silently reduced to the single character `"`, losing the label. This restores the text, adds a verifier fixture, and documents quoting as the way to carry commas and pipes in a link label.

## Problem

`_top_level_mask` blanks quoted spans so operators *inside* a label are never mistaken for link syntax — correct and necessary. But `_edge_operators` then sliced the label out of the `spaced` capture group. Against an all-blank span the non-greedy `.+?` has nothing to anchor on and settles on a single masked character; slicing the *original* text at that one-character offset lands on the closing quote.

So the label became `"`. Silently — no error, no exit 2. The digest then feeds a diagram with a garbage edge label, which is the failure mode that gets noticed last.

Scope, measured before fixing:

| Input | Mermaid 11.17.0 renders | Extractor produced |
|---|---|---|
| `A -- "quoted spaced" --> B` | `quoted spaced` | `"` |
| `A == "thick quoted" ==> B` | `thick quoted` | `"` |
| `A -. "dashed quoted" .-> B` | `dashed quoted` | `"` |
| `A <-- "bidi quoted" --> B` | `bidi quoted` | `"` |
| `A o-- "circ quoted" --o B` | `circ quoted` | `"` |
| `A x-- "cross quoted" --x B` | `cross quoted` | `"` |
| `A -- "a\|b" --> B` | `a\|b` | `"` |

All are valid input — `mermaid.parse()` accepts each one, and the rendered text above is what Mermaid's own renderer draws. Quoting also isn't cosmetic here: `-->\|a\|b\|` can't express a pipe, so the quoted spaced form is the only way to get one into a link label.

Confined to flowchart inline labels. The pipe form (`-->\|"text"\|`), quoted node labels, and the sequence/state/ER grammars take their label after a delimiter, never through this slice — I checked each, all were already correct.

## Solution

Read the label from the whole span between the opening and closing operators instead of from the capture group:

```python
clean_label(text[match.end("opening") : match.start("closing")])
```

The compact branch's group already spans exactly that region, so its behavior is unchanged; the spaced branch gains the surrounding padding, which `clean_label` already strips — along with surrounding quotes, which it was already doing for node labels. No new regex, no new helper.

## Testing

- **Differential test against Mermaid 11.17.0's own renderer**, 33 cases. Before: 8 quoted forms wrong. After: 0 mismatches on every quoted *and* unquoted form.
- **Before/after on the same 33 cases**: 14 changed, every one a quoted label going from `"` to its correct text. No unquoted form changed — including the chained, multidirectional, and `x`/`o`-endpoint cases the existing fixtures cover.
- **New verifier fixture** (`quoted-labeled-links.mmd`) covering all six stroke/direction forms, delimiter-bearing text, and padding. Confirmed it *fails* against the pre-fix extractor (`['"', '"', '"', '"', '"', '"', '"', '"']`) and passes after — it would have caught this.
- **Full CONTRIBUTING gate suite passes**, including `lint-render.py --self-test` and `--all` (147 files rendered, 0 findings) and `claude plugin validate . --strict`.
- Plugin version bumped to 2.6.7 (2.6.6 is taken by my open #151); `verify-plugin-package.py origin/main` passes. Happy to rebase and re-bump if something lands first.

## Not in scope

Two pre-existing divergences I found while testing and deliberately left alone, since neither is caused or worsened by this change:

- `A --- tie --- B` (spaced label on an undirected `---` link) exits 2 both before and after.
- `A -- "has --> arrow" --> B` — Mermaid itself renders this lossily as `has -- arrow`; we now emit the user's literal text. Pathological either way.

Happy to open issues for either if you'd like them tracked.

## Visual proof

Not applicable — no visual output changes. The diff is one extractor expression, a verifier fixture, one documentation sentence, and the version bump.

## Validation gates

- [x] `git diff --exit-code` green after `python3 scripts/build-icons.py` (icons unchanged)

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples (none changed)
- [x] Rendered screenshots are attached for every new/changed diagram variant, or visual proof is marked not applicable

## Related issues

No existing issue — filing under CONTRIBUTING's "small fixes and docs can go straight to a PR". Happy to open one first if you'd rather track it that way.

Heads-up on overlap: #113 edits the same paragraph of `references/import-mermaid.md` (adding grammars to the same sentence), so whichever lands second needs a one-line rebase. It doesn't touch `_edge_operators`, so there's no functional conflict.


cathrynlavery (changes requested):
Thank you @Jenil-1714 — excellent diagnosis and an unusually thorough writeup. I reproduced the bug on current `main`: every quoted spaced label (plain, dotted, thick, and pipe/comma-bearing text) comes back as the single character `"`, silently. Your read-between-the-operators fix is the right shape, the explanatory comment will save the next person real time, and the new `quoted-labeled-links.mmd` fixture in `verify-mermaid-import.py` is exactly what CONTRIBUTING asks for — at your head the full mermaid gate passes, including the compact-multidirectional cases #131 added, so there's no regression on that front.

Two things before this can land:

- The branch is conflicting with `main` again — releases kept landing after your merge of main, and manifests are now at 2.6.11, so 2.6.7 no longer passes `verify-plugin-package.py origin/main`. Please rebase and re-run `python3 scripts/bump-plugin-version.py`.
- Your current head is a merge commit CI never ran on (checks are green on 234a7f3 but absent on bd2d1c5), so after rebasing let CI run on the new head.

Once it's rebased with a fresh bump this is ready to merge from my side. And yes — please do open the two issues you flagged (`A --- tie --- B` exiting 2, and the lossy `has --> arrow` case); both are worth tracking separately.

cathrynlavery:
Thanks @Jenil-1714 — this fix landed on main via #174, preserving your authorship and the exact extractor, regression fixture, and docs changes. Only the stale manifest bump remains here, so I’m closing as superseded.

PR #158, opened 2026-08-28
feat(i18n): put Korean label rules on the always-loaded path

mysoniw:
Continues @etaseoul-jason's #86 on current `main`. That one closed unmerged after your revise request, and #83 has since landed the fallback allowlist it was waiting on, so this sticks to the three things you asked for — the always-loaded guidance, template font availability, and a width budget that survives mixed labels.

## The width budget

#86 measured its own per-script advances (11px per Hangul syllable, 6.5px per Latin character, 3px per space), and that is where the sizing gap came from. `주문 v2.1` is two full-width syllables and five narrow characters; a formula that tallies Hangul, Latin letters, and spaces silently drops `2`, `.`, and `1` and sizes the box for four of the seven.

Rather than patch that formula I removed it and pointed the rules at the model `verify-treemap.py` already implements, the one #121 landed: every Unicode wide or full-width character costs 1em, every other character costs its face's Latin advance (0.60em sans, 0.62em mono), and nonspacing marks cost nothing. It walks characters rather than scripts, so a character class cannot go missing — digits and punctuation included. It also matches what `output-spec.md` already tells importers, so the two surfaces stop disagreeing.

I tried to be careful about what that claims. The rules say `verify-treemap.py` enforces this text width for treemap cell labels; the padding and the round-to-4 are authoring convention, and no other type carries an automatic check. If you would rather it read as a general contract I can widen the wording, but I did not want to promise enforcement that isn't there.

## Changes

- `style-guide.md` gains a Korean labels section — the two fallback stacks, the width budget, the 12px floor, and the register switch for arrow labels, eyebrows, and legend text, which sit at 7–8px mono where Hangul has neither a face nor legibility.
- `SKILL.md` carries one line plus the two Korean faces in its font link, so a plain "draw me a flowchart" sees the rule without loading a reference. The line costs 300 bytes and the link 68, leaving 96 under the 40,000 cap. If that is too tight I can cut it to the pointer and the 12px floor.
- The four templates load Noto Sans KR and Noto Serif KR, and Noto Serif KR joins `--font-serif` so a mixed Latin/Korean title stops falling through to the platform's generic serif and rendering differently on macOS and Windows. Latin glyph selection is unchanged — Instrument Serif still leads.
- Shipped examples keep the shorter link. None of them contains Hangul, so there is nothing there to resolve, and `build-icons.py` output is untouched.
- `output-spec.md` leads the Korean stack with the shipped web font instead of the local families, matching the link, and hands the register and floor rules to `style-guide.md`. The two Korean fixtures in `test-lint-a11y.py` follow the same order.
- `lint-skin.py` allows `noto serif kr`; #83 already covered the sans side.

## Validation

Adversarial cases for the mixed labels you named live in `test-verify-treemap.py` — a version suffix (`주문 v2.1`, 5.0em), a leading Latin acronym (`API 게이트웨이 v2`, 9.2em), full-width parentheses, a decomposed combining mark that exercises the `Mn` branch, and the empty string, plus the mono face's own narrow advance so the two faces cannot silently converge.

The full gate suite from `CONTRIBUTING.md` runs green against current `main` — 39 gates, including `claude plugin validate . --strict`, `lint-skin.py --all --baseline`, `lint-render.py --self-test && --all`, `verify-screenshot-freshness.py`, and the `build-icons.py` regeneration diff. Manifests are at 2.6.8 across Claude, Codex, and Factory.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "diagram-design",
   "description": "Create branded architecture, IT current-state, flowchart, sequence, state machine, ER/data model, timeline, swimlane, quadrant, radar/spider, polar chart (polar/radial lollipop), loop/flywheel, nested, tree, org chart, layer stack, Venn, pyramid/funnel, treemap, bar, line, Gantt and scatter charts, high-level, process, medallion, data flow, DP integration, DP security matrix, sankey, fishbone, Wardley map, kanban, user journey, deployment, dependency graph, UML class, story map, or database schema diagrams as standalone HTML/SVG/PNG. Redraw .drawio/.drawio.png/.drawio.svg or Mermaid .mmd sources at a chosen size/detail; onboard brand tokens from a website; add semantic patterns, callouts, accessible motion, or sketchy/hand-drawn styling.",
-  "version": "2.6.7",
+  "version": "2.6.8",
   "author": {
     "name": "Cathryn Lavery",
     "url": "https://github.com/cathrynlavery"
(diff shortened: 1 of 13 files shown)


mysoniw:
Closing this — it went into main with #160, so there's nothing left here. Thanks for folding it in and for keeping the co-author line.

PR #155, opened 2026-08-25
fix(doctor): quote argument-hint so frontmatter is valid YAML

chck:
## What does this PR do?

`commands/doctor.md` carries `argument-hint: [--strict] [--json]` unquoted. YAML reads `[--strict]` as a flow sequence and then hits a second one on the same line, so the frontmatter is invalid and strict YAML consumers drop the command. `commands/profile.md` and `prompts/doctor.md` already quote the same shape; this makes `commands/doctor.md` consistent with them.

The other three command files start with `<html-file>` / `<drawio-file>` / `<mermaid-file>`, which YAML reads as a plain scalar, so they parse fine and are left alone. `doctor` is the only file that starts with `[`.

Reproduction — installing the plugin with [apm](https://github.com/danielmeppiel/apm) (`apm install cathrynlavery/diagram-design --target claude`) integrates 4 of the 5 commands:

```
|-- 4 commands integrated -> .claude/commands/
[!] [diagram-design] Skipped command doctor.prompt.md: malformed or over-budget
    YAML frontmatter (while parsing a block mapping
      in "<unicode string>", line 2, column 1:
        description: Run one-shot enviro ...
        ^
    expected <block end>, but found '['
      in "<unicode string>", line 3, column 27:
        argument-hint: [--strict] [--json]
                                  ^)
```

Claude Code's own frontmatter parser is lenient, so `/diagram-design:doctor` works there either way — this only affects consumers that parse the frontmatter as strict YAML.

After the change, every file under `commands/` and `prompts/` round-trips through `yaml.safe_load`:

```
ok   commands/doctor.md
ok   commands/export-diagram.md
ok   commands/import-drawio.md
ok   commands/import-mermaid.md
ok   commands/profile.md
ok   prompts/doctor.md
ok   prompts/export-diagram.md
ok   prompts/import-mermaid.md
ok   prompts/profile.md
```

Rebased on `main` after 2.6.11 landed; version bumped to 2.6.12 via `scripts/bump-plugin-version.py` per CONTRIBUTING.

## Visual proof

Not applicable — no visual output changes.

## Validation gates

- [x] `python3 scripts/test-lint-a11y.py`
- [x] `python3 scripts/lint-skin.py --all --baseline` — 147 files, 0 findings
- [x] `python3 scripts/verify-sequence-oauth.py`
- [x] `python3 scripts/verify-drawio-import.py`
- [x] `python3 scripts/verify-mermaid-import.py`
[...]

PR #169, opened 2026-08-31
feat(patterns): add traceable block decomposition semantic pattern

### LINKED ISSUE #170: Semantic pattern: traceable, ID-addressable block decomposition
## The need

Some documentation needs a diagram where each structural element carries a stable, citable identifier — not just a name — so a reader (a compliance review, an audit, IP/patent-support documentation) can point at one specific block and trust it resolves to the same thing later. None of the current 8 semantic patterns or 39 visual types give a block that kind of addressable identity: every diagram shows structure, but nothing lets a reader cite one specific block directly, trace it to its parent, to what it consumes and produces, and to the code that implements it.

## Intended shape

- A new semantic pattern (not a new visual type), routed to the existing **Tree** grammar — Tree already has the right shape for this without any layout change: single parent per node, no cycles, real multi-child breadth.
- The visible surface stays inside the existing density budget: a small ID badge, the block's name, and at most one short in/out port label, using the existing node-box type-tag chip and Geist Mono sublabel primitives. No new SVG grammar.
- The full per-block record (inputs, outputs, constraints, assumptions, an implementation path) lives in `data-block-*` attributes on the node, not on the canvas — following the precedent `type-polar.md`/`type-line.md` already set with their own `data-*` chart contracts, each with a dedicated validator.
- A new `--registry` export mode projects those attributes into a `<basename>.registry.json` sidecar on demand, for anything that needs the full record as structured data rather than a picture.
- A dedicated verifier checks the registry actually coheres as a tree: unique IDs, every parent resolving to a real sibling ID, presence of a name, no cycles in the parent chain.

Happy to discuss routing, naming, or scope here before/alongside a PR.

Hank-tha-Cowdog:
## What does this PR do?

Adds a new semantic pattern, **Traceable block decomposition**, for hierarchical, ID-addressable
block diagrams where each block carries inputs/outputs/constraints/assumptions and a link to the
code that implements it. Routes to the existing Tree type per ADR 0002/0007; visible surface stays
inside the density budget (an ID badge, a name, an optional short port label); the full per-block
record lives in `data-block-*` attributes, projectable on demand into a `.registry.json` sidecar via
a new `--registry` export mode. Full rationale in `docs/adr/0010-block-registry-metadata-contract.md`.

## Visual proof

| Variant | Rendered preview |
|---|---|
| Light | Light variant |
| Dark | Dark variant |
| Full editorial | Full editorial variant |

New example: `example-tree-block-decomposition.html` (+ `-dark`, `-full`), gallery-registered under
a new "Tree · block decomposition" tab (eyebrow 53). A generic payment-gateway decomposition —
deliberately not tied to any real product — chosen so the example reads clearly as illustrative.

## Validation gates

- [x] `git diff --exit-code` green after `python3 scripts/build-icons.py` (icons untouched by this PR)

Also ran the complete gate list from `CONTRIBUTING.md` end to end (all green, including
`lint-render.py --all` and `verify-plugin-package.py --require-no-bump origin/main`) and the pinned
CI validator, `npx @anthropic-ai/claude-code@2.1.229 plugin validate . --strict`.

## Checklist

- [x] Accessible SVG contract satisfied for new/changed examples
- [x] Rendered screenshots are attached for every new/changed diagram variant

## Related issues

Closes #170.

---

Addresses the full round of review on the previous head:

1. **Version collision with `main`** — `main` now bumps versions after merge (ADR 0009, #172), so
   this PR carries no bump: all three manifests match `main` at 2.6.12 and
   `verify-plugin-package.py --require-no-bump origin/main` passes. Current `main` is merged into the
   branch rather than rebased, so the reviewed history is intact.
2. **Issue first** — opened #170 before this push, describing the need and intended shape.
3. **No example exercising the pattern** — shipped light/dark/full variants (see Visual proof
   above), gallery-registered; `verify-block-registry.py --all` now checks 158 files, not 155. The
   badges render each block's `data-block-id` (`PAY-001`, `PAY-001-01`, …); the first cut showed
   category tags there instead, which is § 8's own named anti-pattern. Screenshots re-rendered.
4. **SKILL.md byte headroom took out operative guidance** — restored both sentences (the
   `verify-geometry.py` pointer under rule 6, and "the §6 connector rules never relax" in §11) and
   found the bytes elsewhere (tightened the output-dials intro, the decorative-SVG rule, and the
   duplicate-ID rationale). Net +5 bytes; 39,989 of 40,000.
5. **ADR framing** — reframed the Context section around the general need; dropped the named
   downstream project and patent-filing detail, which now lives in issue #170 instead. The same
   framing is also gone from ADR 0002's amendment line, `semantic-patterns.md` § 8, the worked
   example in `export-registry.md`, and the full variant's headline card; the reference docs now use
   the shipped payment-gateway example throughout. The ADR is `docs/adr/0010-…` since `main` took
   0009 for the version policy.
6. **Housekeeping** — PR title now matches the head commit's conventional-commit shape; both new
   gates (`verify-block-registry.py --all`, `test-verify-block-registry.py`) are registered in
   `.maintainer-policy.json` next to the geometry pair, and documented in `CONTRIBUTING.md`'s gate
   table, combined script, and troubleshooting section.

Also from a second pass over this head:

7. **Verifier hardening** — a blank or whitespace-only `data-block-id` is now a finding, a blank id
   never resolves a blank `data-block-parent`, and a whitespace-only `data-block-name` is caught like
   an empty one; five new adversarial cases cover both polarities.
8. **`export-registry.md` ordering** — the schema note and procedure step 5 gave two different
   definitions of document order; it is now defined once (source order, never re-sorted by parent).


Hank-tha-Cowdog:
Thanks for the review — both flagged issues were real, verified independently, and are fixed as of 654fbba:

1. **`--registry` inheriting the image-export default.** Fixed in `commands/export-diagram.md` and `prompts/export-diagram.md`: a bare `--registry` (no `--svg-only`/`--png-only`) is now explicitly registry-only and never pulls in Playwright.
2. **Parser truncating on a quoted `>`.** Confirmed by reproducing it: a `data-block-*` value containing a literal `>` (valid HTML) positioned before `data-block-id` in the tag caused the entire block to silently vanish from the scan — a real block, clean CI pass. `scripts/verify-block-registry.py`'s tag matcher now treats a quoted span (either quote style) as one unit instead of stopping at the first `>`, and the attribute matcher accepts single quotes, which it previously didn't. Four new regression cases cover both truncation orderings, single quotes, and mixed quoting in one tag.

Also re-ran the full local check suite plus the pinned CI validator (`plugin validate . --strict`) after the fix — all clean.

cathrynlavery (changes requested):
Thanks for the fast, substantive turnaround on the Greptile findings — and for verifying them independently rather than just patching what the bot said. I re-ran the full gate suite at 654fbba: the quote-aware tag matcher is exactly the right fix (a value like `data-block-constraint="output > input"` before `data-block-id` no longer silently drops the block), the four regression cases cover both orderings plus single and mixed quoting, and the registry-only export contract in `commands/export-diagram.md` / `prompts/export-diagram.md` is now unambiguous about never requiring Playwright. That whole cluster is resolved as far as I'm concerned.

A few things still stand before this can merge, one of them new since you branched:

1. **Version bump now collides with main.** PR #155 merged today and moved all three manifests on main to 2.6.12 — so this PR's 2.6.11 → 2.6.12 bump no longer increases, and `verify-plugin-package.py origin/main` fails locally at your head ("must increase relative to origin/main: 2.6.12 -> 2.6.12" for all three manifests). A rebase onto current main plus re-running `scripts/bump-plugin-version.py` (landing at 2.6.13) fixes this mechanically.

2. **Issue first.** CONTRIBUTING asks for an issue before anything non-trivial, and a new semantic pattern + export mode + CI gate + ADR is well past that bar. Please open one describing the need and the intended shape — it can be short, and I'm happy to discuss routing/scope there. This isn't bureaucracy for its own sake; it's how we keep design discussion out of review threads and give future readers a trailhead.

3. **The pattern ships with no example that exercises it.** No file under `skills/diagram-design/assets/` carries `data-block-id`, so `verify-block-registry.py --all` passes vacuously — the gate you built has nothing to guard yet. The PR body mentions a real 3-block Tree you validated against locally; please ship it (light/dark/full variants per the new-surface convention, gallery-registered), along with rendered screenshots in the PR body per the "Visual proof" section of the template. That also gives reviewers something concrete to judge the visible ID badge / port-label treatment on.

4. **SKILL.md byte headroom came out of operative guidance.** Rule 6 lost its "verify with `python3 <repo-root>/scripts/verify-geometry.py <file>`" instruction, and the sentence "The §6 connector rules never relax" is deleted outright. Both are load-bearing: the first is the only in-context pointer from the label-mask rule to its verifier, and the second forecloses an argument we've actually seen agents make. Please restore both and find the bytes elsewhere — happy to point at softer candidates (there's trimmable prose in the size-preset paragraphs) if the cap gets tight.

5. **ADR 0009 framing.** The context section grounds the decision in a downstream project's future patent filing (First Cut V3). ADRs here document *this repo's* rationale — settled policy that future contributors read as precedent. Please reframe the context around the general need (stable, citable block identity with a machine-readable sidecar for compliance/IP-style documentation) and drop the named project and filing; that motivation belongs in your issue or PR description, not settled policy.

6. **Housekeeping.** (a) Conventional-commit PR title, please — your head commit already has the right shape (`feat(patterns): …`); the PR title should match since it becomes the squash subject. (b) The two new CI steps (`verify-block-registry.py --all`, `test-verify-block-registry.py`) should be registered in `.maintainer-policy.json` alongside the geometry pair so the policy manifest stays the source of truth for required gates.

One more note: CI hasn't run on this PR yet (first-contribution approval gate on our side, not anything you did wrong) — I'll approve the run once the rebase lands so the pinned validator and render lint execute against the real head.

The core design here is genuinely good — Tree routing per ADR 0002/0007 is the right call, the three-way visibility split respects the density budget, and the verifier is now the most adversarially-tested script in `scripts/`. The remaining items are all shippable-scope fixes. Looking forward to the next push.

Hank-tha-Cowdog:
@cathrynlavery Thanks for the detailed review — all six addressed as of latest push. Point by point:

1. **Version collision** — rebased onto current `main`, re-ran the bump helper (now 2.6.13), `verify-plugin-package.py origin/main` passes locally.
2. **Issue first** — opened #170 with the need and intended shape.
3. **No example** — shipped light/dark/full variants (`example-tree-block-decomposition*.html`), gallery-registered as "Tree · block decomposition" (eyebrow 53). `verify-block-registry.py --all` now covers 158 files, not 155. Screenshots in the PR body.
4. **SKILL.md byte headroom** — restored both sentences you named (the `verify-geometry.py` pointer under rule 6, and "the §6 connector rules never relax" in §11), and found the bytes elsewhere: tightened the output-dials intro, dropped a redundant clause after the decorative-SVG rule, tightened the duplicate-ID rationale. Net +5 bytes; 39,989 of 40,000.
5. **ADR 0009 framing** — reframed the Context section around the general need (stable, citable block identity + machine-readable sidecar for compliance/IP-style documentation); dropped the named project and filing, which now live in #170 instead.
6. **Housekeeping** — PR title matches the head commit's conventional-commit shape; both new gates registered in `.maintainer-policy.json` next to the geometry pair, and documented in `CONTRIBUTING.md`'s gate table, combined script, and troubleshooting section.

Also ran the complete `CONTRIBUTING.md` gate suite end to end after all of the above (including `lint-render.py --all`, which needed Playwright installed) — all green — plus the pinned CI validator. Let me know if the rebase needs the run-approval re-triggered on your end.

Hank-tha-Cowdog:
@cathrynlavery Pushed 6aa6fd3, which also picks up current main. Beyond the six points already addressed, a second pass over that head found a few things worth fixing before you look again:

1. The example's badges showed category tags (SYS, AUTH, RISK, LDG) while `data-block-id` carried `PAY-001`, `PAY-001-01`, and so on. That is § 8's own "ID badge with no matching `data-block-id`" anti-pattern, and the full variant's headline card already claimed the badge said `PAY-001-02`. The chips now carry the IDs, widened to fit, with node boxes, labels and the density budget unchanged. Screenshots re-rendered.

2. Your point 5 had been applied to the ADR only. The same downstream-project and future-filing framing survived in ADR 0002, `semantic-patterns.md` § 8, the `export-registry.md` worked example, and the full variant's card. All of it now uses the shipped payment-gateway example, and the worked example is a literal excerpt of running `--registry` on it.

3. `verify-block-registry.py` accepted `data-block-id=""` as a valid block, and a blank `data-block-parent` resolved against it. Blank or whitespace-only ids are findings now, a blank id never resolves a parent, and whitespace-only names are caught like empty ones. Five new adversarial cases.

4. `export-registry.md` defined document order twice, once as "depth-first" and once bare. It is now defined once: source order, never re-sorted by parent.

5. Versioning: with #172 on main this PR no longer bumps. All three manifests match main at 2.6.12 and `--require-no-bump` passes. Since main took ADR 0009 for that policy, this PR's ADR is now 0010. main is merged into the branch rather than rebased; happy to rebase if you would rather have a linear history before the squash.

The full CONTRIBUTING gate suite (all 50 commands, including `lint-render.py --all`) and the pinned validator ran green on the merged tree. CI still needs the first-contribution run approval on your side.

cathrynlavery:
I validated exact head 6aa6fd3 against current main: all 52 current commands and both new block-registry gates pass, and all variants have matching IDs, parent links, accessibility metadata, and clean rendered output. No additional code changes are requested. Please resolve the outdated thread and leave this open while maintainer review and the first-contributor CI run are refreshed.

cathrynlavery (changes requested):
The prior six review items are resolved, the 54 maintained commands and eight current-head CI jobs pass, and desktop/mobile/light/dark/full/print proof is clean. One new blocking parser gap remains: verify-block-registry.py still does not parse valid HTML the way a browser does.

On ca983b2, parse_blocks() returns zero blocks for unquoted attributes, attributes with whitespace around “=”, and case-insensitive DATA-BLOCK-* names. It also treats a commented-out block as live and skips a boolean/blank data-block-id instead of reporting it. Because check() treats zero parsed blocks as legal, malformed registries can pass CI silently—the same fail-open class the quoted-“>” fix addressed.

Please replace the regex tag/attribute scanner with browser-equivalent HTML parsing (stdlib html.parser.HTMLParser is sufficient), preserve source-order/line reporting, ignore comments and raw-text contents, and add adversarial tests for unquoted attributes, whitespace around “=”, case-insensitive names, blank boolean attributes, comments, and duplicate-attribute behavior matching the browser. No other changes are requested.

cathrynlavery (approved):
Verified exact head d1876f7 against current main 2.6.16. The HTMLParser implementation closes every reported browser-semantics gap: unquoted, whitespace-separated, and case-insensitive attributes parse; comments, script, and style content stay inert; valueless IDs become blank findings; and duplicate attributes use browser first-wins behavior.

All 54 exact-head policy commands and all eight current hosted CI jobs pass. The reviewed visual assets are byte-identical to the previously inspected desktop/mobile, light/dark, and print/static output; manifests remain unchanged under ADR 0009; and no review threads remain unresolved. Approved.

Hank-tha-Cowdog:
@cathrynlavery Fixed. verify-block-registry.py's regex tag/attribute scanner is replaced with the stdlib html.parser.HTMLParser, which gives the parity you asked for for free: it already lowercases attribute names, tolerates unquoted values and whitespace around `=`, and never calls handle_starttag for tag-like text inside a comment or a `<script>`/`<style>` element, so a commented-out block is never scanned as live.

The one behavior that needed explicit code: a present-but-valueless `data-block-id` (a boolean attribute, no `=` at all) now registers as a block with a blank id — a `find_blank_ids` finding — instead of the attribute being absent from the parsed dict and the whole element vanishing from the scan. And a repeated attribute name now keeps its first value, matching how a browser resolves a duplicate attribute within one tag.

Six new adversarial cases cover unquoted values, whitespace around `=`, case-insensitive `DATA-BLOCK-*` names, the boolean/blank id, the commented-out block, and duplicate-attribute precedence. I checked each one would actually fail under the wrong implementation (last-attribute-wins, a naive regex re-scanning comment text) before trusting it, not just that it passes now.

Source order and line reporting are unchanged; the rest of the checker (duplicates, orphan parents, missing names, cycles) is untouched. Full gate suite green on exact head after merging current main back in.

PR #183, opened 2026-09-05
fix(export): bound the rasterize wait so proxied networks don't hang

### LINKED ISSUE #176: Export recipe hangs behind proxy: goto(load) + networkidle waits on Fonts subrequest
## Setup
- macOS arm64, Python Playwright 1.62.0, bundled Chromium 151 / system Chrome 152
- Egress via proxy (`https_proxy` set); direct net restricted
- Diagram: repo-shipped `assets/example-architecture.html` (self-contained + Google Fonts `<link>`)

## Repro
Follow `references/export.md` Rasterize snippet verbatim:
`page.goto(file://...)` (default `wait_until="load"`) raises `TimeoutError: 30000ms exceeded`.
The page itself is local — the hang is the `fonts.googleapis.com/css2` stylesheet subrequest, which stalls without succeeding or failing fast. Same URL via curl through the same proxy returns 200 in ~3s, so the proxy path works; the browser request stalls.

## Workaround used
`wait_until="domcontentloaded"` + bounded wait with fallback:
```python
page.goto(url, wait_until="domcontentloaded")
try:
    page.wait_for_load_state("networkidle", timeout=15000)
except Exception:
    page.wait_for_timeout(4000)
```
Export then succeeds (webfont may fall back to local stack).

## Suggestion
Note the fallback in the recipe for proxied / slow networks, or prefer `domcontentloaded` + bounded `networkidle` wait as the default snippet.

atomsbaza:
## What does this PR do?

Fixes #176. The rasterize recipe in `references/export.md` used `page.goto(file://...)` with the default `wait_until="load"` followed by an unbounded `page.wait_for_load_state("networkidle")`. Behind a proxy, the page's Google Fonts stylesheet subrequest can stall without succeeding or failing fast, so `load` never fires and the recipe hangs. The snippet now prefers `domcontentloaded`, then waits for `networkidle` with a 15s cap and falls back to a short settle wait — on a stalled webfont the export still succeeds with the local font stack.

## Visual proof

Not applicable — documentation-only change to an export recipe; no diagram, example, or template output changes.

## Validation gates

- [x] `python3 scripts/verify-docs-sync.py` (OK — description hooks, gallery reachability, README tree, reference links, packaged support files, routing surfaces)
- [x] `python3 scripts/test-verify-docs-sync.py` (PASS)
- [x] `python3 scripts/verify-plugin-package.py --require-no-bump origin/main` (OK — Claude, Codex, and Factory 2.6.12 untouched)
- [ ] Remaining gates run by CI on this PR (docs-only change; no icons, examples, or skins touched)

## Checklist

- [x] Generated and source files are consistent (no generated files touched)
- [x] Accessible SVG contract satisfied (no SVG changes)
- [x] Visual proof marked not applicable (no visual output changes)

## Related issues

Closes #176

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,7 @@ Every validation gate below must pass before a PR is ready. They also run automa
 | Every shipped motion template/example | `python3 scripts/verify-motion.py --shipped` |
 | Docs/routing sync (description hooks, gallery, README tree, reference links, strict-bundler support paths, command/prompt surfaces) | `python3 scripts/verify-docs-sync.py && python3 scripts/test-verify-docs-sync.py` |
 | Canonical README screenshots match their example HTML sources and recorded PNG digests | `python3 scripts/verify-screenshot-freshness.py` |
+| Export snippet's stalled-webfont fallback (window.stop on timeout, warning, normal load) | `python3 scripts/test-export-wait.py` (requires Playwright; skips without it) |
 | README WebP previews match their PNGs, manifest, dimensions, and full-size links | `python3 scripts/test-build-readme-thumbs.py && python3 scripts/build-readme-thumbs.py --check` (requires `Pillow==12.1.1`) |
 | Packaged output self-check behaves (pass + adversarial cases) | `python3 scripts/test-self-check.py` |
 | Label masks are never clipped by a node painted after them | `python3 scripts/verify-geometry.py --all` |
@@ -129,7 +130,8 @@ python3 scripts/test-plugin-package.py \
   && python3 scripts/verify-beeswarm.py --all \
   && python3 scripts/test-verify-beeswarm.py \
   && python3 scripts/verify-skin-polarity.py --all \
-  && python3 scripts/test-verify-skin-polarity.py
+  && python3 scripts/test-verify-skin-polarity.py \
+  && python3 scripts/test-export-wait.py
 ```
 
 ### If a gate fails
--- a/scripts/test-export-wait.py
+++ b/scripts/test-export-wait.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+"""Deterministic tests for the export snippet's stalled-load fallback.
+
+Drives the exact python snippet shipped in
+skills/diagram-design/references/export.md against local fixtures:
+
+- a stalled stylesheet request is cancelled with window.stop(), the snippet
+  warns about fallback typography, and capture completes quickly;
+- a normal load captures cleanly with no warning;
+- non-timeout errors are not swallowed by the fallback.
+
+Requires Playwright (and its Chromium); skips otherwise, matching the export
+procedure's own detection step.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import io
+import re
+import sys
+import tempfile
+import threading
+import time
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parent.parent
+EXPORT_DOC = ROOT / "skills/diagram-design/references/export.md"
+
+# Fast-but-deterministic substitutes for the snippet's production waits.
+FAST_IDLE_TIMEOUT = "timeout=1500"
+FAST_SETTLE = "wait_for_timeout(150)"
+STALL_HOLD_SECONDS = 30  # must outlast the test's networkidle timeout
+WARNING_ANCHOR = "fallback typography"
+# The full fallback (fast timeout + settle + capture) must stay far below the
+# production path's 15s networkidle + 4s settle budget.
+FALLBACK_BUDGET_SECONDS = 15.0
+
+
+def load_snippet() -> str:
+    text = EXPORT_DOC.read_text(encoding="utf-8")
+    blocks = re.findall(r"```python\n(.*?)```", text, re.S)
+    if len(blocks) != 1:
+        raise AssertionError(
+            f"expected exactly one python block in {EXPORT_DOC.name}, found {len(blocks)}"
+        )
+    snippet = blocks[0]
+    for anchor in (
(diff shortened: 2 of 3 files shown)


cathrynlavery (changes requested):
The switch to domcontentloaded is useful, but a stalled font still fails after roughly 49 seconds: networkidle times out, the fallback waits, then screenshot performs its own font-readiness wait. Cancel the outstanding load before capture (for example window.stop after the specific Playwright TimeoutError), catch only that timeout, warn that fallback typography was used, and add deterministic stalled-stylesheet/font plus normal-load validation.

atomsbaza:
All four points are addressed in a3cb2143 (amended into the single commit):

1. **`window.stop()` on timeout only** — the fallback branch now calls `page.evaluate("window.stop()")` to cancel the stalled load before capture, triggered exclusively by the `networkidle` wait timing out.
2. **Narrow catch** — `except Exception` replaced with `except PlaywrightTimeoutError`; any other error propagates and fails the export normally.
3. **Fallback warning** — a stderr warning (`webfont request stalled - captured with fallback typography...`) is printed, and the doc now tells the agent to pass it on to the user.
4. **Deterministic tests** — new `scripts/test-export-wait.py` (wired into the CONTRIBUTING validation chain) covers a stalled stylesheet served by a loopback server that never responds, a normal load, and a non-timeout error propagating. It also pins the snippet's anchors: requires `except PlaywrightTimeoutError:` and rejects any broad/bare `except`, so a regression to a swallowing handler fails the gate.

Before: stalled font made the whole capture fail in ~49s (15s timeout + 4s fallback + a second font-readiness wait at screenshot time). After: the stalled case captures in ~3s with the warning; the normal case is unchanged (~1s, no warning).

pr183-evidence-composite

cathrynlavery:
Thanks for addressing the timeout path on head `a3cb2143acefa072fa9962a3f6d395896afdf573`: the snippet now catches only `PlaywrightTimeoutError`, calls `window.stop()`, warns about fallback typography, and the added test covers stalled, normal, and non-timeout cases.

One remaining gap: `scripts/test-export-wait.py` is listed in `CONTRIBUTING.md`, but it is not actually invoked by `.github/workflows/ci.yml`. The current green CI therefore does not provide evidence that this new browser-dependent regression ran. In particular, the test exits successfully with `SKIP` whenever Playwright is absent, while the workflow installs Playwright/Chromium only later for the Linux/Python 3.12 render checks.

Please add an explicit CI step on that same Linux/Python 3.12 leg after **Install Playwright Chromium**, run `python scripts/test-export-wait.py`, and expose its outcome in the execution summary. Then verify the job log shows all three export-wait cases passing rather than a skip. That will turn the deterministic fixture into an enforced gate and satisfy the current review request on this head.

cathrynlavery (changes requested):
Thanks for taking on this frustrating proxy edge case — making export predictable on constrained networks matters. Exact head `a3cb214` remains behind current main, passes 51 of 58 current commands, and retains the unresolved timeout thread. Please cancel the stalled load before continuing, catch only Playwright `TimeoutError` rather than masking unrelated launch/navigation faults, emit a visible fallback warning, and add deterministic tests for both a stalled font request and the normal loaded path. Rebase onto current main without manifest edits, resolve the thread, and run the complete policy table. I appreciate the care you’ve put into this.

atomsbaza:
Done on head `baeb0faf48747b12b54afecc907305b55655f410` (amended into the single commit): `.github/workflows/ci.yml` now invokes `scripts/test-export-wait.py` explicitly.

- The new step **Verify export snippet stalled-load fallback** runs on the ubuntu-latest / Python 3.12 leg, immediately after **Install Playwright Chromium**, so the browser fixture is guaranteed to be present.
- Because the script exits 0 with a `SKIP:` line when Playwright is missing, the step guards against that outcome: a SKIP is escalated to a `::error::` failure, and the step also requires the `All export-wait cases passed.` completion line in the job log — so a green run certifies that the stalled-load, normal-load, and non-timeout-propagation cases all executed, not skipped.
- The outcome is exposed in the CI execution summary table as **Export Snippet Stalled-Load Fallback**.

Verified locally on the same script before pushing: stalled stylesheet cancelled via `window.stop()` in 2.9s with the fallback-typography warning, normal load captured with no warning, and a non-timeout error propagated as `Error` — `All export-wait cases passed.`

cathrynlavery:
Thank you — the hosted regression step now has the right shape and explicitly fails on `SKIP`, which addresses the substantive review gap. Please rebase once onto current `main`, keep the manifests unchanged, and rerun the complete exact-head suite. The Linux/Python 3.12 job should show all three export-wait cases executing and `All export-wait cases passed.` Once that refreshed head is conflict-free and green, I’ll validate and merge it individually.

philyuchkoff (approved):
# Review: fix(export): bound the rasterize wait so proxied networks don't hang — PR #183

@atomsbaza 

This is a solid fix for #176. The current export snippet hangs behind a proxy when a font stylesheet subrequest stalls — `wait_until="load"` blocks forever. Switching to `domcontentloaded` + bounded `networkidle` with a 15s timeout + `window.stop()` on timeout + specific `PlaywrightTimeoutError` catch is the right sequence.

The test suite (`test-export-wait.py`) is thorough — three scenarios: stalled stylesheet (window.stop kicks in, warning fires, capture completes), normal load (no warning), and non-timeout errors propagate. The stall server on localhost with a dedicated handler + deterministic timeout shortening for testing is clean engineering.

Per the maintainer's feedback: the snippet already catches `PlaywrightTimeoutError` specifically (not generic `TimeoutError`), and `window.stop()` cancels the outstanding load before capture. The only remaining concern would be rebasing onto current main — `a3cb214` is behind `dc1ace4` — and the remaining failing CI commands.

LGTM on the logic and gate. Approving — the snippet itself is correct.

PR #196, opened 2026-09-07
fix(export): keep exported SVG font imports in sync

### LINKED ISSUE #195: [Bug]: exported .svg @import requests no CJK face, so exports drop the Korean and Chinese fonts the HTML link ships
### Area

Export (`references/export.md`)

### What happened?

The `@import` that `export.md` step 3 injects into an exported `.svg` carries **none of the four CJK faces** the HTML `<link>` ships. A CJK diagram exports to a `.svg` whose Latin type is web-font-pinned and whose CJK type is not.

On `main` (2.6.15):

```
<link>   Instrument Serif · Geist · Geist Mono · Noto Sans KR · Noto Serif KR · Noto Sans TC · Noto Serif TC
@import  Instrument Serif · Geist · Geist Mono
```

Missing from the export: `Noto Sans KR`, `Noto Serif KR`, `Noto Sans TC`, `Noto Serif TC`.

This is not a request to add Chinese. Korean has shipped in the link far longer than Traditional Chinese has, and the export drops it too — so the export path has been asymmetric with the HTML path for as long as the Korean faces have been there. #186 left it alone deliberately: adding only the Chinese faces would have restored one asymmetry by deepening another.

### Why it stays hidden

`style-guide.md#korean-labels` states the reason the faces are in the link at all:

> the web font resolves before any locally installed one and **the same file renders identically on macOS, Windows, and a reviewer's browser**

An exported `.svg` gets the opposite. It resolves CJK through whatever the viewer has locally, so it looks correct on the machine that produced it and changes on the machine that receives it. The HTML beside it is fine, which is what makes this quiet: the author checks the HTML, sees correct type, and ships the `.svg`.

`export.md` already warns that offline tools substitute typography and recommends the PNG export for pixel-perfect portability. That warning is about tools which don't fetch remote fonts at all. This is different — a modern browser fetching the `@import` correctly still gets no CJK face, because none is requested.

### Steps to reproduce

Build a `.svg` exactly as `export.md` step 3 prescribes, with one CJK label:

```bash
python3 - <<'PY'
import pathlib, re
b = pathlib.Path("skills/diagram-design")
imp = re.search(r"@import url\('([^']+)'\)",
                (b/"references/export.md").read_text(encoding="utf-8")).group(1)
pathlib.Path("repro.svg").write_text(
    f'<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 240 60">'
    f'<defs><style>@import url(\'{imp}\');</style></defs>'
    f'<text x="10" y="34" font-family="\'Geist\', \'Noto Sans TC\', sans-serif" '
    f'font-size="16">請求項比對</text></svg>', encoding="utf-8")
PY
```

Open `repro.svg` in a browser on a machine with no Chinese (or Korean) system font, or in one with a different one than the author's. The Latin runs resolve to Geist from the `@import`; the CJK run does not resolve to any requested face.

The families actually requested by each string:

```bash
python3 - <<'PY'
import pathlib, re
b = pathlib.Path("skills/diagram-design")
fam = lambda u: {p.split(":")[0] for p in u.replace("&amp;","&").split("&") if p.startswith("family=")}
link = re.search(r'href="([^"]*css2[^"]*)"', (b/"assets/template.html").read_text(encoding="utf-8")).group(1)
imp  = re.search(r"@import url\('([^']+)'\)", (b/"references/export.md").read_text(encoding="utf-8")).group(1)
print(sorted(fam(link) - fam(imp)))
PY
# ['family=Noto+Sans+KR', 'family=Noto+Sans+TC', 'family=Noto+Serif+KR', 'family=Noto+Serif+TC']
```

### Environment

`main` @ 2.6.15, macOS, Python 3.14. Also reproduces from an installed plugin copy.

### Suggested fix

Make the exported `@import` request the same families as the shipped `<link>`, with `&amp;` separators as `export.md` already requires (that escaping was fixed in #37, so the string has been maintained before):

```
...family=Geist+Mono:wght@400;500;600&amp;family=Noto+Sans+KR:wght@400;500;600&amp;family=Noto+Serif+KR:wght@400&amp;family=Noto+Sans+TC:wght@400;500;600&amp;family=Noto+Serif+TC:wght@400&amp;display=swap
```

`css2` slices CJK by unicode-range, so a Latin-only diagram downloads nothing extra — the same argument `#korean-labels` already makes for keeping the link short.

A gate would stop the two strings drifting again. The families in `export.md`'s `@import` should be a superset of, or equal to, those in `assets/template.html`; the check is the second snippet above plus an assertion, and it belongs next to `verify-docs-sync.py`.

I'm happy to send the PR if you'd like it — say whether you want the gate in the same one.

HunterKK1424:
Closes #195.

The `@import` that `export.md` step 3 injects requested three Latin faces while `assets/template.html` requested seven. The four missing ones were all the CJK faces, so a Korean or Chinese diagram exported to `.svg` lost its type — quietly, and only on a machine other than the author's.

## The fix

`references/export.md` — the `@import` now requests the same families as the shipped link, with the `&amp;` separators the surrounding prose already requires:

```
…family=Geist+Mono:wght@400;500;600
&amp;family=Noto+Sans+KR:wght@400;500;600&amp;family=Noto+Serif+KR:wght@400
&amp;family=Noto+Sans+TC:wght@400;500;600&amp;family=Noto+Serif+TC:wght@400
&amp;display=swap
```

`css2` slices CJK by unicode-range, so a Latin-only diagram downloads nothing extra — the same argument `#korean-labels` already makes for keeping the link short.

## The gate

`scripts/verify-docs-sync.py` gains `check_export_font_parity`: the families in `export.md`'s `@import` must be a superset of those in `assets/template.html`'s `<link>`.

It lives in `verify-docs-sync.py` rather than a new script on purpose. A new `verify-*.py` would need matching entries in `ci.yml`, the CONTRIBUTING gate table, and `.maintainer-policy.json` — and `test-maintainer-policy.py` asserts those three agree, so the cheap change is the one that reuses a gate already wired into all of them. This is also the same class of check the file already performs: two surfaces that must not drift apart.

## The gate demonstrably catches the reported bug

Reverting only `export.md` to its state on `main` and re-running the gate:

```
FAIL docs sync
  - references/export.md @import omits Noto Sans KR, Noto Sans TC, Noto Serif KR,
    Noto Serif TC, which assets/template.html requests; an exported .svg would
    resolve those scripts through whatever font the viewer happens to have
```

Exit 1 on the pre-fix tree, exit 0 with the fix. A gate that has never been seen to fail is a gate nobody has verified, so it seemed worth doing before opening this.

## Why the parity direction is superset, not equality

`template.html` is the source of truth for what ships. An export may legitimately request more (a future terminal or print variant), but never less — dropping a face is the failure this is about. Equality would also break the moment anything is added to the export side alone.

## Scope

Two files. No new script, no CI wiring, no manifest versions touched. Reads as a patch to me; label it otherwise if you disagree.

### DIFF
--- a/scripts/verify-docs-sync.py
+++ b/scripts/verify-docs-sync.py
@@ -556,6 +556,53 @@ def check_manifest_descriptions(errors: list[str], root: Path) -> None:
                     )
 
 
+def font_families(url: str) -> set[str]:
+    """The `family=` parameters a Google Fonts css2 URL actually requests."""
+    return {
+        part.split(":", 1)[0]
+        for part in url.replace("&amp;", "&").split("&")
+        if part.startswith("family=")
+    }
+
+
+def check_export_font_parity(errors: list[str], root: Path) -> None:
+    """The exported SVG must request every face the shipped HTML link does.
+
+    The two strings live in different files and drifted apart once already: the
+    CJK faces reached assets/template.html but never the @import in export.md,
+    so a Korean or Chinese diagram exported to .svg silently lost its type. That
+    failure only shows up on a machine other than the author's, which is exactly
+    the case the faces are in the link to prevent.
+    """
+    template = root / "skills/diagram-design/assets/template.html"
+    export = root / "skills/diagram-design/references/export.md"
+    for path in (template, export):
+        if not path.is_file():
+            errors.append(f"font-parity surface is missing: {path.name}")
+            return
+
+    link = re.search(r'href="([^"]*fonts\.googleapis\.com[^"]*)"',
+                     template.read_text(encoding="utf-8"))
+    imported = re.search(r"@import url\('([^']+)'\)",
+                         export.read_text(encoding="utf-8"))
+    if not link or not imported:
+        errors.append(
+            "could not locate the font link in assets/template.html or the "
+            "@import in references/export.md"
+        )
+        return
+
+    missing = sorted(font_families(link.group(1)) - font_families(imported.group(1)))
+    if missing:
+        names = ", ".join(name.removeprefix("family=").replace("+", " ")
+                          for name in missing)
+        errors.append(
+            f"references/export.md @import omits {names}, which "
+            f"assets/template.html requests; an exported .svg would resolve "
+            f"those scripts through whatever font the viewer happens to have"
+        )
+
+
 def main() -> int:
     errors: list[str] = []
     check_description(errors)
@@ -581,6 +628,7 @@ def main() -> int:
     )
     check_line_dark_skin(errors, LINE_DARK_EXAMPLE.read_text(encoding="utf-8"))
     check_routing_surfaces(errors, ROOT)
+    check_export_font_parity(errors, ROOT)
     if errors:
         print("FAIL docs sync")
         for error in errors:
@@ -590,7 +638,8 @@ def main() -> int:
         "OK docs sync: description hooks, gallery reachability, README tree, "
         "reference links, asset citations, packaged support files, routing surfaces, "
         "manifest descriptions, Factory install contract, type-count routing, "
-        "High-Level invariants, onboarding trust boundary, Line dark-skin contract"
+        "High-Level invariants, onboarding trust boundary, Line dark-skin contract, "
+        "export font parity"
     )
     return 0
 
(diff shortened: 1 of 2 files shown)


cathrynlavery (approved):
Verified exact head 17e8ab9 against current main: all 52 policy commands and all nine hosted checks pass, there are no review threads, and plugin manifests remain unchanged under ADR 0009. Reverting only the export @import makes the docs-sync gate fail with all four missing CJK families; restoring this change passes. The implementation is correct and focused.

PR #197, opened 2026-09-07
feat(profiles): support repository-local profile snapshots

### LINKED ISSUE #159: [Feature]: Repository-local Diagram Design profiles
### What are you trying to show?

Currently, a project can select a profile with a `.diagram-design` marker:

```
profile: foo
```

but that slug resolves only to:

```
~/.diagram-design/profiles/foo.md
```

That works per-user, but it makes repository-shared branded diagrams hard to reproduce: every contributor must manually install the same profile in their home directory.

Having this would let teams commit a complete brand/profile definition with the repository, making generated diagrams reproducible across machines and CI without requiring manual home-directory setup.

### Proposed solution

Add support for repository-local profiles, for example:

```
.diagram-design/
  current-profile
  profiles/
    foo.md
```

Where `.diagram-design/current-profile` uses the same grammar as the current marker file:

```
profile: foo
```

Resolution could be:

1. If `.diagram-design/current-profile` exists:
    - read profile: `<slug>`
    - resolve `.diagram-design/profiles/<slug>.md`
    - optionally fall back to `~/.diagram-design/profiles/<slug>.md`

2. Else, keep supporting the current `.diagram-design` file behavior:
    - read profile: `<slug>`
    - resolve `~/.diagram-design/profiles/<slug>.md`

### Compatibility

- Keep the existing `.diagram-design` file format working.
- New directory format is opt-in.
- Slugs should keep the existing validation rules.
- Repository profile files should still be treated as untrusted data, not instructions.

### Acceptance criteria

- A repo can contain `.diagram-design/current-profile` and `.diagram-design/profiles/<slug>.md`.
- Diagram generation uses that repo-local profile.
- Existing `~/.diagram-design/profiles/<slug>.md` behavior remains unchanged.
- Missing or malformed repo-local profiles fail visibly instead of silently choosing another profile.

### Scope

Not sure yet — help me scope it

### Design-system fit

- [x] I've checked the anti-patterns and 'When not to use' — this still seems worth drawing
- [x] I'm willing to contribute the implementation or pair with a maintainer

sdhilip200:
Closes #159

## Summary
- Support directory-backed `.diagram-design` markers with `current-profile`.
- Prefer committed repository-local profiles, then fall back to the home library.
- Keep repository profiles untrusted, structurally checked, and outside the `save` and `update` flows.

## Validation
- `python3 scripts/test-verify-docs-sync.py`
- `python3 scripts/verify-docs-sync.py`
- `python3 scripts/verify-semantic-motion.py --markdown-only`
- `python3 scripts/test-maintainer-policy.py`
- `python3 scripts/verify-plugin-package.py --require-no-bump origin/main`


cathrynlavery (changes requested):
The repository-local profile shape matches #159, and exact head `f40da53` passes all 54 policy commands and all hosted jobs. One contract conflict remains: `default` is reserved for the pristine shipped profile, but the directory-marker rule resolves `.diagram-design/profiles/default.md` first. That lets repository content silently replace the built-in default while reporting that the default profile is active.

Please make `profile: default` bypass repository-local lookup and resolve only the protected home-library default—or explicitly redesign the reserved-slug contract everywhere—add an adversarial docs-sync fixture for a local `default.md`, and resolve the existing P1 thread.

sdhilip200:
Updated the resolver contract so profile: default always uses the protected home-library default and never selects .diagram-design/profiles/default.md. Added an adversarial docs-sync case for that local override path, rebased onto current main, and verified docs sync, semantic routing, maintainer policy, and plugin packaging.

cathrynlavery (changes requested):
Thank you for the thoughtful follow-through here — repository-local profiles are a genuinely useful addition. The repository-local profile direction is good and exact head `29af797` passes all 58 local gates on current main. One security boundary is still missing. The contract treats `.diagram-design/current-profile` and `.diagram-design/profiles/<slug>.md` as untrusted repository data, but validates only the slug; a committed symlink can redirect either read outside the project before the structural check runs. Please require regular, non-symlink marker/profile files or resolve the real path and prove it remains beneath the intended project directory, and add red/green contract fixtures for symlink escape plus the normal repository-local fallback. Then rerun the complete suite on the new exact head. I appreciate the care you’ve put into this.

PR #210, opened 2026-09-10
fix(import): allow whitespace in compact dotted-link labels

### LINKED ISSUE #209: [Bug]: mermaid_extract rejects `-.label with space.->` — #80's compact-label fix excludes whitespace
### Area

Mermaid import

### What happened

`mermaid_extract.py` exits 2 with `malformed edge at line N` on a dotted link whose label contains a space and no padding spaces around it. Mermaid renders this form fine — it appears in production docs that render on GitHub.

This looks like an incomplete fix of #80. That issue asked for the compact label forms to be recognised and named `-.text.->` explicitly among them. The resulting `compact` branch does handle `-.text.->`, but only while the label is a single word.

Three variants isolate the trigger to whitespace alone:

| Source | Exit |
|---|---|
| `A -.next candidate.-> B` | **2** |
| `A -. next candidate .-> B` | 0 |
| `A -.retry.-> B` | 0 |

**Why:** `_edge_operators` (`skills/diagram-design/scripts/mermaid_extract.py:611-619`) alternates a `spaced` branch requiring `\s+` padding against a `compact` branch of `[^\s|<>]+?`. With no padding after `-.`, only `compact` can match, and it stops at the first space — so the closing `.->` is never reached and the line is rejected.

The comment at :608 states the whitespace exclusion is deliberate, and for dash-delimited links it is clearly right: it keeps `A --o B --> C` two links rather than one link with the label `o B `. But `-.text.->` is delimited by the dots, not by whitespace, so the restriction is guarding against an ambiguity that form does not have.

### Second effect, which is the more disruptive one

`--diagram N` does not let you route around it. Every block in a Markdown file is parsed before the requested one is selected, so a single bad edge in block 0 makes **every** diagram in that file unreachable:

```
$ python scripts/mermaid_extract.py playbook.md --diagram 1
mermaid_extract: malformed edge at line 66     # line 66 is in block 0
```

That turns "one diagram is unavailable" into "this file is unusable". On a document with several diagrams, one legacy edge blocks all of them.

### Steps to reproduce

```bash
printf 'flowchart LR\n  A[Start] --> B[End]\n  A -.next candidate.-> B\n' > repro.mmd
python skills/diagram-design/scripts/mermaid_extract.py repro.mmd
# mermaid_extract: malformed edge at line 3   (exit 2)

printf 'flowchart LR\n  A[Start] --> B[End]\n  A -. next candidate .-> B\n' > ok.mmd
python skills/diagram-design/scripts/mermaid_extract.py ok.mmd
# parses (exit 0)
```

### Suggested fix

Allow the `compact` branch to admit whitespace when the closing operator is dot-delimited (`\.-+[>xo]`), leaving the dash and equals forms under the existing rule. The dots bound the label in the common case; a label containing a literal dot (`-.v1.2 ready.->`) stays ambiguous, but non-greedy matching resolves it the same way it already does for the spaced branch.

Separately, deferring per-block parse errors until after `--diagram N` selection would stop one bad edge from taking out an entire multi-diagram file.

`scripts/verify-mermaid-import.py` passes on this version, so nothing currently covers the case — its "documented shape families and edge forms normalize" check looks like the natural home for a regression test.

### Environment

Windows 11 ARM64, Python 3.14.4, Claude Code. diagram-design v2.6.21 (`562dbdf`), installed via the Claude Code plugin marketplace.

### Validation gates

Ran `python scripts/verify-mermaid-import.py` — exits 0 on an unmodified checkout, i.e. it does not catch this.

Kumaresan-AI-Engineer:
mermaid_extract's compact-form regex rejected any labeled dotted link whose text contained a space (`A -.next candidate.-> B`), because the compact branch forbade whitespace uniformly across dash, equals, and dotted operators. That restriction is load-bearing for dash/equals forms (their closing operator is just more dashes, so whitespace would be ambiguous), but the dotted form's closer always opens with a literal `.`, so it doesn't need it.

Split the combined regex into a dash/equals pattern (unchanged) and a dotted pattern whose compact label may contain internal whitespace, and run both before the existing unlabeled-arrow fallback.

Fixes #209.

## Visual proof

| Variant | Rendered preview |
|---|---|
| Light | |
| Dark | |
| Full editorial | |

## Validation gates

- [ ] `python3 scripts/test-lint-a11y.py`
- [ ] `python3 scripts/lint-skin.py --all --baseline`
- [ ] `python3 scripts/verify-sequence-oauth.py`
- [ ] `python3 scripts/verify-drawio-import.py`
- [ ] `python3 scripts/verify-mermaid-import.py`
- [ ] `git diff --exit-code` green after `python3 scripts/build-icons.py` (if icons changed)
- [ ] Docs updated in the same PR where behavior changed

## Checklist

- [ ] No new entries added to `scripts/lint-skin-baseline.txt`
- [ ] Generated and source files are consistent (extractor ↔ verifier ↔ reference ↔ command)
- [ ] Accessible SVG contract satisfied for new/changed examples
- [ ] Rendered screenshots are attached for every new/changed diagram variant, or visual proof is marked not applicable
- [ ] Code of Conduct respected


cathrynlavery (changes requested):
The current head correctly accepts whitespace in compact dotted-link labels and adds focused coverage, but `Fixes #209` promises both reported effects. `--diagram N` still parses every Mermaid block before selecting N, so a malformed unselected block can still abort extraction. Please either implement selection-before-parse with an adversarial regression, or change the closing linkage to `Refs #209` and leave #209 open with a follow-up for the second effect. Rerun the full policy suite on the resulting head.

Kumaresan-AI-Engineer:
PR fix/mermaid-dotted-label-whitespace addresses the first effect reported here  the compact-form regex rejecting a dotted link's label when it contains whitespace (`A -.next candidate.-> B`). That's fixed by splitting the dash/equals and dotted operator patterns so only the dotted form's compact label allows internal whitespace, with a regression fixture covering the whitespace, bidirectional, and circle-marker cases.

The second effect — `--diagram N` still parsing every block before selection, so a malformed unselected block aborts extraction of the whole file — is **not** addressed by that PR. It's a bigger change than it looks: the digest header lists every diagram's kind/node/edge counts up front (so you can browse a multi-diagram file before picking one), which means making unselected-block failures non-fatal requires the header to degrade gracefully per block instead of just omitting them, `_fail()` to become catchable per-block rather than raising `SystemExit` directly, and both the Markdown and `--json` output shapes to carry per-block error state.

Updated the commit to `Refs #209` instead of `Fixes #209` so this stays open for that second effect, and reran the full validation gate suite on the resulting head — all green (two unrelated pre-existing failures on `main` aside: the `SKILL.md` byte-cap overage and a screenshot-freshness/autocrlf line-ending mismatch, neither touched by this branch).

cathrynlavery (changes requested):
Thanks for narrowing this cleanly and for being explicit about the second effect in the commit. The focused dotted-label fix at exact head `e7a5d7f` is sound, but the PR body still says `Fixes #209` even though the commit correctly says `Refs #209` and the reported `--diagram N` parse-before-selection failure remains. Change the PR closing linkage to `Refs #209` (leaving the issue open for effect two), or implement selection-before-parse with an adversarial multi-block regression. Then rebase onto current main, keep manifests unchanged, and rerun all 58 policy commands; this head is behind and currently fails the no-bump compatibility gate. I appreciate the care you’ve put into this.

cathrynlavery:
The focused dotted-label fix remains the right scope, and I appreciate the clear explanation of the second issue. One visible mismatch remains: the PR body still says `Fixes #209`, although the parse-before-selection failure is intentionally left for follow-up. Please change that linkage to `Refs #209`, rebase once onto current `main` without manifest edits, and rerun the complete current policy suite. Once the refreshed exact head is conflict-free and green, I’ll validate and merge this fix individually while leaving #209 open.

PR #234, opened 2026-09-17
fix(mermaid): allow whitespace in compact dotted edge labels

### LINKED ISSUE #209: [Bug]: mermaid_extract rejects `-.label with space.->` — #80's compact-label fix excludes whitespace

rufushsu9987:
## Summary

Fixes #209.

`mermaid_extract` rejected compact dotted edge labels that contain whitespace (`-.next candidate.->`) while accepting the spaced form (`-. next candidate .->`) and single-word compact form (`-.retry.->`). The `#80` compact-label branch used `[^\s|<>]+?`, which stops at the first space, so the closing `.->` was never reached.

Dot-delimited labels are bounded by the dots, so they do not share the dash ambiguity that requires excluding whitespace (`A --o B --> C` must stay two links). This PR:

1. Adds a `compact_dotted` alternative that allows whitespace when the closing operator is dot-delimited (`(?=\.-)`), leaving dash/equals compact labels under the existing no-whitespace rule.
2. Extends `verify-mermaid-import.py` with a regression covering `-.next candidate.->`, padded/spaced and single-word dotted forms, a label with an interior version-like dot (`-.v1.2 ready.->`), and the chained `F --o G --> H` ambiguity guard.

## Test plan

- [x] `python3 scripts/verify-mermaid-import.py` (all Mermaid import gates passed)
- [x] Manual repro from #209: `A -.next candidate.-> B` now exits 0 and retains label `next candidate`
- [ ] CI Lint & Verify matrix


cathrynlavery:
Thanks for adding stronger adversarial coverage. #210 is the earlier implementation, and this head still has an unresolved mixed-chain parser failure. I’m consolidating on #210’s separated-pattern approach, which completed the earlier correctness review, and closing this duplicate.

PR #235, opened 2026-09-17
fix(export): carry CSS and namespace defs IDs for standalone SVG

### LINKED ISSUE #202: SVG export drops class-based styling — 25 of 156 shipped examples export as black boxes
### Area

Export (SVG / PNG)

### What happened?

The SVG export procedure in `skills/diagram-design/references/export.md` extracts the `<svg>` node and injects only a Google Fonts `@import`. It never carries the diagram's own CSS rules.

That is fine for the diagrams that put every fill, stroke and font in inline presentation attributes (`example-architecture.html` and friends). It silently breaks every diagram that styles itself through CSS classes declared in the page `<style>` block — `.station`, `.hub`, `.node-name`, `.ring`, `.spoke`, `.sublabel`, `.arrow-label`, etc. Once the wrapper is dropped, those rules are gone: every shape falls back to `fill: black` and every `<text>` to the default 16px, so the exported SVG is a stack of black boxes with text spilling out of them.

**25 of the 156 shipped `example-*.html` / `template*.html` files (16%) are affected**, spanning 7 diagram types plus the motion template:

```
example-loop{,-dark,-full,-terminal}.html
example-process{,-dark,-full}.html
example-medallion{,-dark,-full}.html
example-data-flow{,-dark,-full}.html
example-dp-integration{,-dark,-full}.html
example-dp-security-matrix{,-dark,-full}.html
example-it-state{,-dark,-full}.html
example-policy-trace-animated.html
example-queue-animated.html
template-motion.html
```

Since `assets/template.html` is the documented starting point for a new diagram and the loop family is the 2.0 flagship, a user following the happy path can very reasonably end up with a class-styled diagram and then a broken export.

The same gap applies to inlining: `export.md` says the prefixed `<title>`/`<desc>` IDs are what make "multiple exported SVGs safe to inline in the same page", which reads as an endorsement of inlining, but a class-styled figure inlined that way is unstyled for the same reason.

### Steps to reproduce

1. Follow the SVG export procedure in `references/export.md` against a class-styled diagram, e.g. `skills/diagram-design/assets/example-loop.html`: extract the first `<svg>…</svg>` block, add the Google Fonts `@import` inside `<defs>`, prepend the XML prolog, write `example-loop.svg`.
2. Open `example-loop.svg` in a browser.
3. Every station, the hub and the ring render with default `fill: black`; node names and sublabels render at the default 16px and overflow their boxes. Compare with `example-architecture.html`, whose export is correct because its styling is inline.

Reproduced by hand and then via a script on a v2.6.17 checkout (`2724fd2`).

### Environment

diagram-design 2.6.17 (`2724fd2`), macOS 15, Chromium 141, Python 3.13.3

### Validation gates

`self_check.py` passes on the source HTML both before and after, so it does not catch this — it validates the document it is given, and the exported SVG is a different document that is never re-checked. `verify-geometry.py` and `lint-skin.py` likewise run against the source file.

### Suggested fix

Two options, roughly in order of effort:

1. **Carry the rules into the SVG.** Copy the page's rules into a `<style>` inside `<defs>`, scoped under an id placed on the `<svg>` so nothing leaks when several figures share one host document, and re-scope `:root` custom properties onto that same id so the figure keeps its own tokens. Drop the page-level `svg { min-width: … }` rule, which must not follow a fragment. Skip page chrome (`body`, `.frame`, `h1`, `.eyebrow`, `*`).
2. **Add a gate.** A fragment that still carries `class=` but has no `<style>` is always broken — that is a one-line check and would have caught this class of bug at authoring time.

I implemented option 1 plus that check locally to unblock myself and can turn it into a PR if you'd like it — happy to follow whatever shape you prefer, including just the doc fix plus the gate.

### LINKED ISSUE #203: export.md's "safe to inline" guarantee misses <defs> IDs — shared marker/pattern IDs collide across figures
### Area

Export (SVG / PNG)

### What happened?

`references/export.md` line 23 states:

> The SVG-only export keeps the source `<title>` and `<desc>` with the diagram. Their per-diagram and per-variant prefixed IDs are what make multiple exported SVGs safe to inline in the same page without one figure resolving to another figure's accessible name.

The accessible-name half of that is true and works well. But the guarantee reads broader than it is: `<defs>` IDs are not namespaced, and every generated diagram declares the same ones. The `dots` pattern is in essentially every diagram, and the markers are shared verbatim across types:

```
<pattern id="dots" …>
<marker id="arrow" …>
<marker id="arrow-accent" …>
<marker id="arrow-link" …>
<marker id="arrow-soft" …>
```

Inlining four figures in one page therefore produces four `id="dots"`, three `id="arrow"`, three `id="arrow-accent"` nodes and so on. That is invalid HTML, and more importantly `url(#arrow)` resolves to whichever declaration came first in document order — so figures 2, 3 and 4 are all rendered with figure 1's markers.

While every figure uses the same skin the fills happen to be identical, so nothing looks wrong and the bug is invisible. It becomes visible as soon as the page mixes skins, which are both documented, supported workflows:

- a light figure next to its `-dark` variant (side-by-side comparison, or a page with both themes),
- two figures generated under different client profiles (`references/profiles.md`),
- a `-terminal` variant next to a normal one.

In those cases the later figures silently take the earlier figure's marker colours. The arrowheads stop matching their own strokes and there is no error anywhere.

### Steps to reproduce

1. Export or extract the `<svg>` node from `assets/example-architecture.html` and from `assets/example-architecture-dark.html`.
2. Inline both into one HTML document, light first.
3. Inspect the dark figure: its `url(#arrow)` and `url(#arrow-accent)` references resolve to the light figure's markers, so its arrowheads render in the light skin's `#4f5d75` / `#eb6c36` instead of the dark skin's values, against dark paper.
4. `document.querySelectorAll('[id="arrow"]').length` returns 2.

### Environment

diagram-design 2.6.17 (`2724fd2`), macOS 15, Chromium 141

### Validation gates

None of the gates cover this: they each validate a single standalone document, where the IDs are unique and correct. The defect only exists in the composed host page, which no gate sees.

### Suggested fix

Either narrow the claim or make it true.

- **Narrow it:** say explicitly that inlining several figures requires namespacing the `<defs>` IDs, and that the guarantee covers accessible names only.
- **Make it true:** in the export procedure, prefix every referenceable `<defs>` ID (`marker`, `pattern`, `linearGradient`, `radialGradient`, `filter`, `clipPath`, `mask`, `symbol`) with the file slug and rewrite the matching `url(#…)` references. Rewriting longest-id-first avoids `arrow-accent` being clipped by the `arrow` rule.

Related to #202, which is the other half of what breaks when a figure is moved out of its page. I have both implemented locally and am happy to open a PR for either or both.

rufushsu9987:
## What does this PR do?

Fixes SVG export for class-styled diagrams and for inlining several figures in one page.

1. **#202** — Exporting a class-styled diagram as a bare `<svg>` dropped the page `<style>` rules, so shapes fell back to black fills. The packaged helper `skills/diagram-design/scripts/export_svg.py` now copies diagram CSS into a scoped `<style>` inside `<defs>` (chrome selectors like `body` / `.frame` / bare `svg { min-width }` are dropped; `:root` tokens rebind to `#<slug>-root`).
2. **#203** — Shared `marker` / `pattern` / gradient IDs (`arrow`, `dots`, …) collided when several exports were inlined. The helper prefixes referenceable defs IDs with the file slug and rewrites `url(#…)` / `href="#…"` longest-id-first.

`references/export.md` documents the helper as the preferred SVG path and narrows the old “safe to inline” claim so it covers defs IDs, not only accessible names. A class-without-`<style>` gate refuses broken fragments.

## Visual proof

Not applicable — export recipe / helper change; no diagram, example, or template source changes. Verification is via the new unit tests exporting shipped assets (e.g. `example-loop.html`, light+dark architecture pair).

## Validation gates

- [x] `python3 scripts/test-export-svg-standalone.py` (PASS — 8 tests)
- [x] `python3 scripts/verify-docs-sync.py` (OK)
- [x] `python3 scripts/test-verify-docs-sync.py` (PASS)
- [x] `python3 scripts/verify-plugin-package.py --require-no-bump origin/main` (OK — versions untouched)
- [x] `python3 scripts/test-maintainer-policy.py` (OK)
- [ ] Remaining gates run by CI on this PR

## Checklist

- [x] Generated and source files are consistent (helper ↔ export.md ↔ SKILL/command/prompt ↔ docs-sync packaging list ↔ CI)
- [x] Accessible SVG contract satisfied (no example SVG changes)
- [x] Visual proof marked not applicable (no visual output source changes)

## Related issues

Fixes #202
Fixes #203


cathrynlavery:
Thank you for the quick follow-up on the exporter review. I rechecked exact head `7bf5d0f6`: the fonts-only stylesheet no longer satisfies the class-style gate, carried CSS is XML-escaped, the exported document is parsed as XML, all nine focused exporter tests pass, and `git diff --check origin/main...HEAD` is clean.

The implementation is worth keeping open. Please rebase once onto current `main`, keep the manifests unchanged under ADR 0009, and run the full hosted matrix on the resulting exact head. With conflict-free status, resolved threads, and current-head green checks, I’ll do the final review and merge it individually.

PR #240, opened 2026-09-20
docs(skill): worst-case text metrics under font substitution; unattended path for style gate

YahyaZekry:
## Context

These two additions to `skills/diagram-design/SKILL.md` come from a production run — a multi-diagram book-dossier where the skill rendered six diagrams end-to-end. Three of the run's four first-round judge failures traced to the same root: optimistic text-width estimation under font substitution.

## Changes (SKILL.md only, no code)

**1. Text metrics under font substitution** (added to the arrow-label rules, §6):

- Budget **≥0.55× font-size per character** for Geist Mono labels (~5px/char at 9px). The intuitive ~4.2px/char estimate overflowed its mask/chip on real renders — twice in one run.
- `rsvg`/`inkscape` previews cannot fetch the Google Fonts the export injects, so a preview renders in a substituted font. Geometry must survive that substitution; generous padding is the defense, and `scripts/verify-geometry.py` is the check (the script ships in the repo — worth noting in-skill since packaged installs made it easy to miss).

**2. Unattended path for the style gate** (§0):

- The first-run gate says "pause and ask the user." In an autonomous session that stalls the run. This adds: if the user is not reachable, take the explicit-default path **(e)**, note the choice beside the deliverable, and move on. Interactive behavior is unchanged.

Both edits are additive and in the existing voice; no existing rules are modified.

## Verification

- The sizing rule (≥0.55× font-size/char) eliminated all text-width judge failures on the run's second round and is derived from that pass/fail evidence.
- `verify-geometry.py` runs offline (no network), so it validates geometry under the same substituted-font conditions as rsvg previews.


YahyaZekry:
Thanks @greptile-apps — addressed in ffbaf6c:

**1. Minimum Width Is Too Small — fixed, you were right.** The 0.55em figure sat below the repo's own calibrated contract (0.60em sans / 0.62em mono, `style-guide.md`), which was self-defeating. The paragraph now references the calibrated budget as the baseline and adds **≥0.7em/char for mono labels under substitution**, framed as margin on top of a budget that "fits Geist Mono exactly" and therefore carries no slack.

**2. Verifier Cannot Check Text Fit — fixed, you were right.** The paragraph now states plainly that `verify-geometry.py` catches masks clipped by later-painted nodes, not text fit, and that the substitution budget "is yours to hold" — matching the style guide's own framing ("no other type carries an automatic check").

**3. Default Style Bypasses Gate — wording strengthened, design defended.** The remedy proposed ("stop and obtain the required style information") is definitionally unavailable in the scenario the paragraph governs — an unattended run has no one to obtain it from, which is why the path exists. What the comment correctly exposes is that my wording let the default pass too quietly: the run now must **state beside the deliverable that the shipped default skin was used and can be restyled on request**. Not silent — recorded, surfaced, and reversible, which is the honest version of the gate's intent when nobody can answer.

YahyaZekry:
CI fix pushed: the additions tripped the repo's own 40,000-byte SKILL.md cap in `verify-semantic-motion --markdown-only` (40,665B). Both blocks are now compressed to the budget (39,974B) with every point intact — calibrated 0.62em baseline, ≥0.7em/char substitution margin, verifier scope, surfaced default-skin note. Local run of the battery: semantic-motion, a11y, and skin lint all pass.

